PR #3, opened 2022-11-25
Further development of M3/lioncore

dslmeinte:
- Refactoring
- adding constraints:
  * an Annotation can't be the type of a Link
  * no cycles from inheritance
- various code level improvements


src/m3/constraints.ts
```diff
@@ -12,19 +19,37 @@ export const issuesMetamodel = (metamodel: Metamodel): Issue[] =>
...
                 const nonDerivedFeatures = t.allFeatures().filter(({derived}) => !derived)
                 if (nonDerivedFeatures.length > 0) {
                     const isPlural = nonDerivedFeatures.length > 1
-                    return [
-                        {
-                            location: t,
-                            message: `The features of a ConceptInterface must all be derived, but the following feature${isPlural ? `s` : ``} of ${t.simpleName} ${isPlural ? `are` : `is`} not: ${nonDerivedFeatures.map(({simpleName}) => simpleName).join(", ")}.`
-                        }
-                    ]
+                    issue(`The features of a ConceptInterface must all be derived, but the following feature${isPlural ? `s` : ``} of ${t.qualifiedName()} ${isPlural ? `are` : `is`} not: ${nonDerivedFeatures.map(({simpleName}) => simpleName).join(", ")}.`)
```
ftomassetti, line 42: Note: this part of having derived features only in ConceptInterface is a bit strange to me, but this is a problem for that decision, not for this implementation which is aligned with the decision we made during the meetings

src/m3/constraints.ts
```diff
@@ -12,19 +19,37 @@ export const issuesMetamodel = (metamodel: Metamodel): Issue[] =>
...
                 }
             }
-            return []
+
+            if (t instanceof Link) {
+                if (t.type instanceof Annotation) {
```
ftomassetti, line 47: This makes me wonder if it should be renamed in t.type, as these two should probably have the same name:

```
Annotation "*" -- "1" AbstractConcept: target
Link "*" -- "1" AbstractConcept: type
```

src/m3/functions.ts
```diff
@@ -20,6 +30,9 @@ export const isPlural = (multiplicity: Multiplicity): boolean =>
 
 type Typed = Link | Property
 
+/**
+ * @return The type of the given {@link Feature}
+ */
 export const type = (feature: Feature): FeaturesContainer | Datatype | typeof unresolved =>
```
ftomassetti, line 36: Does this indicate that we need some abstraction to represent these three elements?

src/m3/functions.ts
```diff
@@ -40,35 +53,53 @@ export const relationsOf = (metamodelElement: MetamodelElement): Link[] =>
         : []
 
 
-export type Comparer<T> = (l: T, r: T) => number
+export const containeds = (thing: M3Concept): M3Concept[] => {
```
ftomassetti, line 56: Here a comment could help to clarify the role of this. Should this be renamed to `containedElements`?

src/m3/functions.ts
```diff
@@ -40,35 +53,53 @@ export const relationsOf = (metamodelElement: MetamodelElement): Link[] =>
...
-        if (thing instanceof Metamodel) {
-            thing.elements.forEach(visit)
-        }
-        if (thing instanceof FeaturesContainer) {
-            thing.features.forEach(visit)
-        }
-        if (thing instanceof Enumeration) {
-            thing.literals.forEach(visit)
-        }
+export type ConceptType = Concept | ConceptInterface
```
ftomassetti, line 82: Does this indicate we have a missing abstraction?

src/m3/test/constraints.test.ts
```diff
@@ -0,0 +1,94 @@
...
+    Containment,
+    Metamodel,
+    Multiplicity,
+    Property,
+    Reference
+} from "../types.ts"
+import {issuesMetamodel} from "../constraints.ts"
+
+
+Deno.test("constraints (lioncore)", async (tctx) => {
```
ftomassetti, line 18: I think it would be important to guarantee compatibility with the boring but established Node.js

src/m3/test/constraints.test.ts
```diff
@@ -0,0 +1,94 @@
...
+
+    await tctx.step("check that a ConceptInterface only has derived features", async () => {
```
ftomassetti, line 20: It would be nice to find a way to share the tests between the different implementations

src/m3/test/constraints.test.ts
```diff
@@ -0,0 +1,94 @@
...
+        const issues = issuesMetamodel(metamodel)
+        assertEquals(issues.length, 1)
+        const {location, message} = issues[0]
+        assertEquals(location, conceptInterface)
+        assertEquals(message, `The features of a ConceptInterface must all be derived, but the following feature of metamodel.conceptInterface is not: property.`)
+    })
+
+    await tctx.step("check that an Annotation is not the type of a Containment", async () => {
+        const metamodel = new Metamodel("metamodel")
+        const annotation = new Annotation(metamodel, "@nnotation")
```
ftomassetti, line 36: Maybe we need rules for identifiers, and @ could be forbidden

src/utils/cycles.ts
```diff
@@ -0,0 +1,62 @@
+export type DependenciesFunction<T> = (t: T) => T[]
+
+
+/**
+ * Compute the dependency order of the given `things`, computing dependencies with the given `dependenciesFunction`.
```
ftomassetti, line 5: Could you provide an example of where this could be used? I think that would help clarify the goal of this as it is a bit abstract

src/utils/test/cycles.test.ts
```diff
@@ -0,0 +1,151 @@
+import {
+    assertEquals,
+    assertFalse
+} from "https://deno.land/std@0.160.0/testing/asserts.ts"
+
+import {cycleWith, dependencyOrderOf} from "../cycles.ts"
+
+
+const collatz = (n: number) => n % 2 === 0 ? n/2 : (3 * n + 1)/2
```
ftomassetti, line 9: What is the purpose of this?

ftomassetti (approved):
I do not see anything blocking, I just added a few comments for things that are not super clear to me and considerations on the M3 that emerged by looking at the code

src/m3/functions.ts
dslmeinte: I'll add a comment. “elements” is not entirely appropriate (I think) as `EnumerationLiteral`s are not `MetamodelElement`s.

src/m3/functions.ts
dslmeinte: That could be, but OTOH: sum types are so cheap in TypeScript that I'm fine with not having the abstraction explicitly in the M3 until it's clear that it's missing in other/all contexts as well.

src/m3/test/constraints.test.ts
dslmeinte: A wrapper for Mocha (which is the most boring and established unit test framework in the Node.js ecology) exists: https://deno.land/x/deno_mocha@0.3.0
But I still don't haven't figured how to have sources that are Deno- and Node.js-compatible at the same time.

src/m3/test/constraints.test.ts
dslmeinte: Yes! For that we'd probably need a language-**a**specific description of the tests...maybe in the form of a model? ;)

src/m3/test/constraints.test.ts
dslmeinte: I'm generally pretty relaxed on this, as I think it's the job of a code generator to map strings to valid identifiers, but a couple of constraints should be there, like:

- Identifiers should be unique (within the namespace)
- An identifier should be a non-empty, non-whitespace-only string

src/utils/test/cycles.test.ts
dslmeinte: Purely for testing the algorithmic stuff - I'll see if I can replace/remove/simplify it.

src/m3/constraints.ts
dslmeinte: Indeed: I'm pretty sure it's something that we explicitly decided for.

src/m3/constraints.ts
dslmeinte: Given that we're going to remove annotations anyway (for now), I guess this problem disappears into thin air :)

PR #4, opened 2022-11-28
Propagate decisions w.r.t. M3 to TypeScript impl.

dslmeinte:
Decisions taken 25-11-2022.

### DIFF
--- a/diagrams/metametamodel-gen.puml
+++ b/diagrams/metametamodel-gen.puml
@@ -20,7 +20,7 @@ abstract class Datatype extends MetamodelElement
 class Enumeration extends Datatype implements NamespaceProvider
 
 abstract class Feature extends NamespacedEntity {
-  multiplicity: Multiplicity
+  optional: boolean
   derived: boolean
 }
 
@@ -28,20 +28,15 @@ abstract class FeaturesContainer extends MetamodelElement implements NamespacePr
   allFeatures(): List<Feature>
 }
 
-abstract class Link extends Feature
+abstract class Link extends Feature {
+  multiple: boolean
+}
 
 class Metamodel implements NamespaceProvider {
   qualifiedName: String
 }
 
 abstract class MetamodelElement extends NamespacedEntity
-
-enum Multiplicity {
-  Optional
-  Single
-  ZeroOrMore
-  OneOrMore
-}
 
 interface NamespaceProvider {
   namespaceQualifier(): String
--- a/diagrams/metametamodel-gen.puml
+++ b/diagrams/metametamodel-gen.puml
@@ -66,8 +66,6 @@ Concept "*" -- "*" ConceptInterface: implements
 
 ConceptInterface "*" -- "*" ConceptInterface: extends
 
-Containment "*" -- "0..1" Containment: specializes
-
 Enumeration "1" o-- "*" EnumerationLiteral: literals
 
 FeaturesContainer "1" o-- "*" Feature: features
@@ -80,8 +78,6 @@ Metamodel "*" -- "*" Metamodel: dependsOn
 NamespacedEntity "*" -- "1" NamespaceProvider: container
 
 Property "*" -- "1" DataType: type
-
-Reference "*" -- "0..1" Reference: specializes
 
 
 @enduml
--- a/diagrams/metametamodel-gen.puml
+++ b/diagrams/metametamodel-gen.puml
@@ -24,6 +24,8 @@ abstract class FeaturesContainer extends MetamodelElement implements NamespacePr
   allFeatures(): List<Feature>
 }
 
+' primitive type: "JSON"
+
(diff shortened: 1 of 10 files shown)


diagrams/metametamodel-gen.puml
```diff
@@ -3,10 +3,6 @@
 ' qualified name: "lioncore"
 
 
-class Annotation extends FeaturesContainer {
```
ftomassetti, line 6: Ok, I forgot about dropping Annotations

diagrams/metametamodel-gen.puml
```diff
@@ -15,32 +11,31 @@ class ConceptInterface extends FeaturesContainer
...
+  optional: boolean
   derived: boolean
 }
 
 abstract class FeaturesContainer extends MetamodelElement implements NamespaceProvider {
   allFeatures(): List<Feature>
 }
 
-abstract class Link extends Feature
+' primitive type: "JSON"
```
ftomassetti, line 27: This comment is not clear to me

diagrams/metametamodel-gen.puml
```diff
@@ -50,23 +45,23 @@ abstract class NamespacedEntity {
...
+class PrimitiveType extends DataType
 
 class Property extends Feature
 
 class Reference extends Link
 
 ' primitive type: "String"
 
-class Typedef extends Datatype
+class Typedef extends DataType
```
ftomassetti, line 56: I think we agreed on having Typedefs, however I cannot see them in the last version of the diagram I can see here https://github.com/LIonWeb-org/organization/blob/main/lioncore/metametamodel.svg

We should figure out which ways to go and make it consistent

diagrams/metametamodel-gen.puml
```diff
@@ -50,23 +45,23 @@ abstract class NamespacedEntity {
...
 
 ' primitive type: "boolean"
```
ftomassetti, line 58: Also this comment is not clear

src/m3/test/self-definition.ts
```diff
@@ -35,14 +40,14 @@ namespaceProvider.havingFeatures(namespaceProvider_namespaceQualifier)
 
 const namespacedEntity = new Concept(lioncore, "NamespacedEntity", true)
 
-const namespacedEntity_simpleName = new Property(namespacedEntity, "simpleName", Multiplicity.Single)
+const namespacedEntity_simpleName = new Property(namespacedEntity, "simpleName")
     .ofType(stringDatatype)
 
-const namespacedEntity_qualifiedName = new Property(namespacedEntity, "qualifiedName", Multiplicity.Single)
+const namespacedEntity_qualifiedName = new Property(namespacedEntity, "qualifiedName")
```
ftomassetti, line 46: Do we have derived features? Here I cannot see them: https://github.com/LIonWeb-org/organization/blob/main/lioncore/metametamodel.svg

ftomassetti (commented):
Just adding a few comments

diagrams/metametamodel-gen.puml
dslmeinte: It's just a result of the primitive types being defined in the self-definition of the M3/lioncore. Because PlantUML doesn't have a 1st-class construct for primitive types I just emit them as comments.

diagrams/metametamodel-gen.puml
dslmeinte: Same thing as above: a consequence of a primitive type def. in the self-definition of M3/lioncore. You can ignore it.

diagrams/metametamodel-gen.puml
dslmeinte: I'll check with that location.

src/m3/test/self-definition.ts
dslmeinte: Right, they've been removed there. But that makes this diagram not definable in M3/lioncore itself. I'll have to think about that a bit.

ftomassetti: Maybe we can leave them for now and mark them as "Experimental"?

ftomassetti:
(I cannot see the button to mark the PR as approved, but please consider it so)

dslmeinte:
The “Approve” button is a little bit hidden: you click the “Review change” button, then select “Approve”, and then click “Submit review” (- buttons may have different but equivalent wording).

diagrams/metametamodel-gen.puml
dslmeinte: I've removed `Typedef`.

src/m3/test/self-definition.ts
dslmeinte: I'll see whether I do the marking with an annotation or something otherwise (JS doesn't really have annotations in the way that Java/the JVM has).

ftomassetti (approved):
+1

PR #6, opened 2022-12-22
(de-)serialization, import from Ecore, and other improvements

dslmeinte:

PR #17, opened 2022-12-26
Implement (de-)serialization

dslmeinte:
Supercedes PR #6, breaking it up into a couple of “atomic” commits which can be reviewed in sequence.

### DIFF
--- a/diagrams/library-gen.puml
+++ b/diagrams/library-gen.puml
@@ -1,0 +1,46 @@
+@startuml
+hide empty members
+
+' qualified name: "library"
+
+
+class Book {
+  title: String
+  pages: int
+}
+
+class GuideBookWriter extends Writer {
+  countries: String
+}
+
+class Library {
+  name: String
+}
+
+class SpecialistBookWriter extends Writer {
+  subject: String
+}
+
+' primitive type: "String"
+
+class Writer {
+  name: String
+}
+
+' primitive type: "boolean"
+
+' primitive type: "int"
+
+
+
+' relations:
+
+Book "*" -- "*" Writer: author
+
+Library "1" o-- "*" Book: books
+
+
+legend
+  <#LightGray,#LightGray>| <#Orange>Disputed |
+end legend
+@enduml
--- a/diagrams/metametamodel-gen.puml
+++ b/diagrams/metametamodel-gen.puml
@@ -1,7 +1,7 @@
 @startuml
 hide empty members
 
-' qualified name: "lioncore"
+' qualified name: "LIonCore"
 
 
 class Concept extends FeaturesContainer {
@@ -50,7 +50,9 @@ abstract class NamespacedEntity {
 
 class PrimitiveType extends DataType
 
-class Property extends Feature
+class Property extends Feature {
+  disputed: boolean
+}
 
 class Reference extends Link
 
--- a/models/README.md
+++ b/models/README.md
@@ -1,0 +1,7 @@
+# README
+
+This directory contains serializations of models in various formats, m.n. the LIonWeb JSON and Ecore XML formats:
+
+* [`library.json`](./library.json): the serialization of the "library" metamodel as [constructed here](../src/m3/test/library.ts).
+* [`lioncore.json`](./lioncore.json): the serialization of the LIonCore/M3 metamodel from its [self-definition](../src/m3/self-definition.ts).
+
--- a/models/lioncore.json
+++ b/models/lioncore.json
@@ -1,0 +1,661 @@
+[
+  {
+    "type": "jj0Y6KYlHt3_qyFDFAx4CA",
+    "id": "31_xAp4rmNVj6Sdahgjeuw",
+    "properties": {
+      "97hsg4Fg3ZitV5FaapRLVg": "LIonCore"
+    },
+    "children": {
+      "1CVT9BH6pWQYU-BZwhppIw": [
+        "yNnK5As--S0BU9Z5CRhqew",
+        "KWfgmX9puPLiNWvWt173Ow",
+        "jj0Y6KYlHt3_qyFDFAx4CA",
+        "FjVUqRRcRRVpZJ_kMQiv2A",
+        "sSN4rvZY6fmq_AqQDAWkWw",
+        "jNyncSWEXyHd08N63CNBBA",
+        "6yDwqIa_ff2-3uABS2hkdg",
+        "nmukD4wNFWVb8ydcjml0SQ",
+        "Bco7VDOeCNdpKX6NKjmXQw",
+        "omtd90jug1c7NnEPjDeELA",
+        "8zwB8soE_0a7rDVPgyxaRQ",
+        "0fOWUa0E9f_pcohEvCa0vw",
+        "-EGokr9qbz6C-K8qdK_PHA",
+        "DKh_5NU9dZ5DmwvDq5attg",
+        "eE2jWdhtKexIOmsTlHJw0A",
+        "vlj29UukiDdjShXmDo_NHw",
+        "PRBb11NZ0hUJ72Ka0RnKRA",
+        "FRb6vvEatAkTf-QU5s6kCA",
+        "EkfEtW80jz64ge2mNjSwRg",
+        "CbxGl35oEY2rF8CLnX2GpA"
+      ]
+    }
+  },
+  {
+    "type": "jNyncSWEXyHd08N63CNBBA",
+    "id": "yNnK5As--S0BU9Z5CRhqew",
+    "properties": {
+      "EvXCQydomgm9DUb1IwkgRg": "NamespacedEntity",
+      "j1tZfW_pl9Cu2NO3TUw4oA": true
+    },
+    "children": {
+      "fSSOzJzcuK14nt-E3K2WTg": [
+        "EvXCQydomgm9DUb1IwkgRg",
+        "1nGRxGQ7owAMjLRgXCVzCA",
+        "YShLQYnrhxLmCOPBqp0oLw"
+      ]
+    },
+    "references": {
+      "noALNU7IxFfZDvrEL840Ig": [],
+      "wUjeOwaFMJPJX5IzUj6WWA": []
+    }
+  },
+  {
+    "type": "8zwB8soE_0a7rDVPgyxaRQ",
+    "id": "EvXCQydomgm9DUb1IwkgRg",
+    "properties": {
+      "EvXCQydomgm9DUb1IwkgRg": "simpleName",
+      "dAp79oxccBi-9j6BrQcZLQ": false,
+      "EE8DYkLRyQzaqyxFqlbOAw": false,
+      "XzYEKJ6QzlZuqKPtMXK_kQ": false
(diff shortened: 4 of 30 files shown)


dslmeinte:
*Notice*: I'll be force-pushing some fixes to this branch in the next couple of minutes!

dslmeinte:
> _Notice_: I'll be force-pushing some fixes to this branch in the next couple of minutes!

done

dslmeinte:
(And the ones after are much smaller.)

diagrams/library-gen.puml
```diff
@@ -0,0 +1,46 @@
+@startuml
```
ftomassetti, line 1: This file is generated, so I assume it does not need to be reviewed

diagrams/metametamodel-gen.puml
```diff
@@ -1,7 +1,7 @@
 @startuml
```
ftomassetti, line 1: This file is generated, so I assume it does not need to be reviewed

models/README.md
```diff
@@ -0,0 +1,7 @@
+# README
+
+This directory contains serializations of models in various formats, m.n. the LIonWeb JSON and Ecore XML formats:
```
ftomassetti, line 3: It seems to me that the directory contains JSON files but not XML files

models/lioncore.json
```diff
@@ -0,0 +1,661 @@
+[
+  {
+    "type": "jj0Y6KYlHt3_qyFDFAx4CA",
+    "id": "31_xAp4rmNVj6Sdahgjeuw",
+    "properties": {
+      "97hsg4Fg3ZitV5FaapRLVg": "LIonCore"
```
ftomassetti, line 6: Hard-code human readable IDs (I think @enikao suggested them in another discussion) could make the file more readable but also help get compatibility with other implementation (e.g., the Java one) easier

schemas/README.md
```diff
@@ -0,0 +1,7 @@
+# README
+
+This directory contains JSON Schemas:
+
+* [`generic-serialization.schema.json`](./generic-serialization.schema.json): schema that captures the LIonWeb serialization JSON format _generically_.
+* [`lioncore.serialization.schema.json`](./lioncore.serialization.schema.json): schema that captures the LIonWeb serialization JSON format _specifically_ for serialized metamodels (i.e., instances of LIonCore/M3).
```
ftomassetti, line 6: Do we want to have both or is the generic one enough? Was this a point that we discussed?

src/id-generation.ts
```diff
@@ -0,0 +1,61 @@
...
+export const sequentialIdGen = (): IdGenerator => {
+    let num = 0
+    return () => `${++num}`
+}
+
+
+const defaultHashAlgorithm = "SHA256"
+
+export type IdGenOptions = {
+    doNotCheckForUniqueData?: boolean
```
ftomassetti, line 23: Perhaps we could add comments to indicate which default values will be used, if the properties are not specified

src/id-generation.ts
```diff
@@ -0,0 +1,61 @@
...
+            throw new Error(`expected data for hashing`)
+        }
+        if (checkForUniqueData) {
+            if (datas.indexOf(data) > -1) {
+                throw new Error(`duplicate data encountered: "${data}"`)
+            }
+            datas.push(data)
+        }
+        const hash = createHash(algorithm)
+            .update(salt + data)
```
ftomassetti, line 49: What is the advantage of using salt here? I am very ignorant of this topic, so I may miss something obvious. I thought that salt was useful to make more difficult to decrypt something, but I think this is not a goal for us in this context

diagrams/library-gen.puml
dslmeinte: (In a couple of PRs, I'll add generation of Mermaid diagrams, so you could see these diagrams directly on GitHub.)

src/m3/deserializer.ts
```diff
@@ -0,0 +1,200 @@
...
+} from "./types.ts"
+import {SingleRef} from "../references.ts"
+import {Id, Node} from "../types.ts"
+import {SerializedNode} from "../serialization.ts"
+import {metaConcepts, metaFeatures} from "./self-definition.ts"
+
+
+export const deserializeMetamodel = (serializedNodes: SerializedNode[]): Metamodel => {
+
+    const metamodelSerNode = serializedNodes.find(({type}) => type === metaConcepts.metamodel.id)
```
ftomassetti, line 24: should we check there is exactly one?

src/m3/deserializer.ts
```diff
@@ -0,0 +1,200 @@
...
+
+    // make a map id -> serialized node:
+    const serializedNodeById: { [id: Id]: SerializedNode } = {}
+    serializedNodes.forEach((node) => {
+        serializedNodeById[node.id] = node
+    })
+
+    const deserializedNodeById: { [id: Id]: Node } = {}
+
+    const constructMemoised = (serNode: SerializedNode, parent?: M3Concept): M3Concept => {
```
ftomassetti, line 37: I do not understand where there is the memoization: construct is always called, even if deserializedNodeById contains an entry already

src/m3/deserializer.ts
```diff
@@ -0,0 +1,200 @@
...
+        const node = construct(serNode, parent)
+        deserializedNodeById[node.id] = node
+        return node
+    }
+
+    const mapConstructMemoised = <T extends Node>(parent: M3Concept) =>
```
ftomassetti, line 43: The role of this element is not clear to me. A comment could help

src/m3/deserializer.ts
```diff
@@ -0,0 +1,200 @@
...
+        (id: Id): T =>
+            constructMemoised(serializedNodeById[id], parent) as unknown as T
+
+    const referencesToInstall: [node: Node, featureName: string, refId: Id][] = []
+
+    const construct = ({type, id, properties, children, references}: SerializedNode, parent?: M3Concept): M3Concept => {
+        switch (type) {
+            case metaConcepts.concept.id: {
+                const {
+                    [metaFeatures.namespacedEntity_simpleName.id]: simpleName,
```
ftomassetti, line 53: I am unfamiliar with this TS construct. I am not sure if this is commonly known or it needs an explanation

src/m3/functions.ts
```diff
@@ -26,7 +28,17 @@ import {flatmapNonCyclingFollowing} from "../utils/recursion.ts"
  * @return The type of the given {@link Feature}
  */
 export const type = (feature: Feature): FeaturesContainer | Datatype | typeof unresolved =>
-    (feature as Link | Property).type
+    (feature as (Link | Property)).type
+
+
+export const isRealProperty = (feature: Feature): feature is Property =>
```
ftomassetti, line 34: In this case real stands for "non derived", correct?
If this is the case I would say that the name is not obvious. Perhaps isNotDerivedProperty could be more obvious

src/m3/functions.ts
```diff
@@ -26,7 +28,17 @@ import {flatmapNonCyclingFollowing} from "../utils/recursion.ts"
...
+    feature instanceof Property && !feature.derived
+
+export const isRealContainment = (feature: Feature): feature is Containment =>
```
ftomassetti, line 37: Same here

src/m3/functions.ts
```diff
@@ -26,7 +28,17 @@ import {flatmapNonCyclingFollowing} from "../utils/recursion.ts"
...
+    feature instanceof Containment && !feature.derived
+
+export const isRealReference = (feature: Feature): feature is Reference =>
```
ftomassetti, line 40: Same here

ftomassetti (approved):
+1 I cannot find anything obviously wrong, but I am not enough skilled in TS to provide valuable feedback

diagrams/metametamodel-gen.puml
dslmeinte: It should actually be compared to the non-generated file in the `organization` repo. Unfortunately, that comparison can't be automated and still needs a human eye because of the “disputed” meta-features that are still in.

src/m3/functions.ts
dslmeinte: Correct. I have a branch/PR lined up that documents these things, and then indeed replaces `Real` with `NonDerived`.

src/m3/deserializer.ts
dslmeinte: It's a JS construct even, and semantically equivalent to: `obj[<property name>] = <value>`.

src/id-generation.ts
dslmeinte: I have a branch/PR lined up in which I do exactly that.

schemas/README.md
dslmeinte: I don't think we discussed it. It serves first and foremost as a check that serialization also behaves correctly in view of a specific metamodel. The generic JSON Schema is rather...permissive, so I think there's value in a specific JSON Schema. E.g., someone could generate type definitions from a metamodel-specific JSON Schema.

models/lioncore.json
dslmeinte: That's a good idea. I made issue #19 for this, as I think this requires a bit of discussion.

models/README.md
dslmeinte: In a couple of PRs, there will also be XML files ;) I'll leave the doc as-is, otherwise I'll have to revert it twice.

src/m3/deserializer.ts
dslmeinte: (I think it's common enough in {J|T}S world.)

src/m3/deserializer.ts
dslmeinte: Agreed: pushing a commit that documents this code.

src/m3/deserializer.ts
dslmeinte: In general: no. A serialized model could contain multiple roots, AFAIA. If we want to constrain this for (serializations of) metamodels we should make that explicit.

src/m3/deserializer.ts
dslmeinte: I fixed that in the commit I just pushed, but even then, `constructMemoised` will be called only once per serialized node. It's good to stick to the memoisation pattern. Otherwise, I would change the name, but also remove the check.

src/m3/functions.ts
dslmeinte: > +1 I cannot find anything obviously wrong, but I am not enough skilled in TS to provide valuable feedback

That makes it all the more important that the code is readable and understandable for non-TS experts as well.

src/m3/functions.ts
dslmeinte: (see above)

src/id-generation.ts
dslmeinte: The advantage is to be able to make hashed IDs unique across models even if you e.g. hash a qualified name. Qualified names (within metamodels) might be the same across models, and would produce the same ID without a salt. Of course, the salt would have to be different... Maybe we can remove this again when we've solved issue #19 .

PR #20, opened 2023-01-02
Implement a (partial) Ecore importer

dslmeinte:


models/README.md
```diff
@@ -2,6 +2,11 @@
 
 This directory contains serializations of models in various formats, m.n. the LIonWeb JSON and Ecore XML formats:
 
+* [`library.ecore`](./library.ecore): Ecore XML file with the “library” Ecore metamodel - copied from [here in the `lioncore-java` repository](https://github.com/LIonWeb-org/lioncore-java/blob/master/emf/src/test/resources/library.ecore).
+* [`library.ecore.json`](./library.ecore.json): a serialization in JSON format of the result of parsing the `library.ecore` using Deno's XML functionality.
```
ftomassetti, line 6: Is it useful to commit this file?

src/m3/ecore/importer.ts
```diff
@@ -0,0 +1,119 @@
...
+import {
+    asArray,
+    EClassifier,
+    EcoreXml,
+    EStructuralFeature
+} from "./types.ts"
+import {ConceptType} from "../functions.ts"
+
+
+const refPrefix = "#//"
```
ftomassetti, line 20: Is this the prefix only for local references? Can we have references that are not local? If so, should we call it "localRefPrefix"?

src/m3/ecore/importer.ts
```diff
@@ -0,0 +1,119 @@
...
+        ? typeDescriptor.substring(refPrefix.length)
+        : typeDescriptor
+
+export const asLIonCoreMetamodel = (ecoreXml: EcoreXml): Metamodel => {
+    const ePackage = ecoreXml["ecore:EPackage"]
+    // TODO  an Ecore XML can contain multiple EPackage-s
+    const factory = new MetamodelFactory(ePackage["@name"], hashingIdGen({ checkForUniqueHash: true }))
+
+    // TODO  obtain from a built-in, imported M3 instance
+    const stringDatatype = factory.primitiveType("String")
```
ftomassetti, line 32: Yes, we should somehow share the built-in metamodel among the different implementations

src/m3/ecore/importer.ts
```diff
@@ -0,0 +1,119 @@
+import {
```
ftomassetti, line 1: Would it make any sense to use some of the EMF libraries for JS/TS? 
For example: https://emfjson.github.io/ecore.js/ or https://github.com/crossecore/ecore-typescript?

src/m3/ecore/types.ts
```diff
@@ -0,0 +1,56 @@
...
+    "@upperBound": string   // number
+    "@eType": string    // `#//${targetType.name}`
+    "@containment": boolean
+}
+
+export const textAsEcoreXml = (data: string): EcoreXml =>
+    parse(data, {emptyToNull: false, reviveNumbers: false}) as unknown as EcoreXml
+
+
+export type UndefOrTOrTs<T> = undefined | T | T[]
```
ftomassetti, line 44: What about ZeroToManyT?

src/m3/ecore/importer.ts
```diff
@@ -0,0 +1,119 @@
...
+                    .ofType(convertEDataType(feature["@eType"]))
+                if (feature["@lowerBound"] === "0") {
+                    property.isOptional()
+                }
+                return property
+            }
+            case "ecore:EReference": {
+                const link = (
+                        feature["@containment"]
+                            ? factory.containment(container, name)
```
ftomassetti, line 69: What I do not understand is what happens in case of a reference to a Classifier we have not yet unserialized: do we create an unresolved reference which is solved later?

ftomassetti:
My main doubt is the fact that no existing libraries for EMF are used. Was that an explicit choice? Would it make sense to consider them?

models/README.md
dslmeinte: Which one do you mean?:

- `library.ecore` is needed to test the importer.
- `library.ecore.json` is Git-ignored, so not committed.

src/m3/ecore/types.ts
dslmeinte: Or `AnyTs`?

dslmeinte:
> My main doubt is the fact that no existing libraries for EMF are used. Was that an explicit choice? Would it make sense to consider them?

That certainly would make sense. Not doing so is a semi-explicit choice: I already knew how to parse XML in Deno, know the Ecore XML format quite well, and hadn't checked yet whether any of the existing JS-EMF libs were already Deno-ready. Using one (provided it has types) would mean that we don't need `src/ecore/types.ts`, and could hopefully use a few more niceties and built-in support for non-local references, so it'd definitely be worth it.

I guess [ecore.js](https://emfjson.github.io/ecore.js/) would be the prime target. Unfortunately, it has a couple of dependencies: `underscore` and `sax`. I'll try and see how far I can get.

src/m3/ecore/importer.ts
dslmeinte: It works because I first convert all EClassifiers but don't convert their EStructuralFeatures yet. I'll swap the functions around to try and make this a bit clearer.

dslmeinte:
> I guess [ecore.js](https://emfjson.github.io/ecore.js/) would be the prime target. Unfortunately, it has a couple of dependencies: `underscore` and `sax`. I'll try and see how far I can get.

Yeah, I didn't get very far yet. Are you OK with me making an issue for migrating to `ecore.js` (or something similar, as long as it's working)?

dslmeinte:
> > I guess [ecore.js](https://emfjson.github.io/ecore.js/) would be the prime target. Unfortunately, it has a couple of dependencies: `underscore` and `sax`. I'll try and see how far I can get.
> 
> Yeah, I didn't get very far yet. Are you OK with me making an issue for migrating to `ecore.js` (or something similar, as long as it's working)?

Haha, ecore.js itself says: “XMI support is provided in ecore.js, but it is limited and does not support all the XMI specification.” So maybe it's less useful to use ecore.js than I thought...

ftomassetti:
I just wanted to make sure we considered ecore.js & similar: if they have been considered and discarded this makes sense to me. I would just suggest adding a comment about that, so that others with the same doubts in the future can understand the rationale behind this choice

ftomassetti (approved):

PR #25, opened 2023-01-13
Use qualified names as IDs for LIonCore/the M3

dslmeinte:


dslmeinte:
Note: I might change some things here and there, and then force-push.

models/lioncore.json
```diff
@@ -1,661 +1,661 @@
 [
   {
-    "type": "jj0Y6KYlHt3_qyFDFAx4CA",
-    "id": "31_xAp4rmNVj6Sdahgjeuw",
+    "type": "LIonCore_M3_Metamodel",
+    "id": "LIonCore_M3",
     "properties": {
-      "97hsg4Fg3ZitV5FaapRLVg": "LIonCore"
+      "LIonCore_M3_Metamodel_qualifiedName": "LIonCore.M3"
```
enikao, line 6: is the dot on the right-hand side on purpose?

models/library-imported-from-ecore.json
```diff
@@ -19,234 +19,234 @@
...
     "id": "Ei9m_HbdmEYg_EwMLLVZ71ERRBZyXH8GHVyOVia8Sqg",
     "properties": {
-      "EvXCQydomgm9DUb1IwkgRg": "title",
-      "dAp79oxccBi-9j6BrQcZLQ": false,
-      "EE8DYkLRyQzaqyxFqlbOAw": false,
-      "XzYEKJ6QzlZuqKPtMXK_kQ": false
+      "LIonCore_M3_NamespacedEntity_simpleName": "title",
+      "LIonCore_M3_Feature_optional": false,
+      "LIonCore_M3_Feature_derived": false,
+      "LIonCore_M3_Property_disputed": false
```
enikao, line 68: what is `Property.disputed`?

schemas/lioncore.serialization.schema.json
```diff
@@ -1,7 +1,7 @@
 {
   "$schema": "https://json-schema.org/draft/2020-12/schema",
-  "$id": "LIonCore-serialization",
-  "title": "Serialization format specific to LIonCore",
+  "$id": "LIonCore.M3-serialization",
```
enikao, line 3: dot on purpose (also next line)?

enikao (approved):

models/lioncore.json
dslmeinte: Yup: I thought the dot was OK for use in qualified name fields, and subsequently qualified names get their dots replaced by underscores to get their IDs.

schemas/lioncore.serialization.schema.json
dslmeinte: See 1st response.

models/library-imported-from-ecore.json
dslmeinte: It indicates for a property whether that property is “disputed” in the sense that
it's not in the agreed-on version of M3/LIonCore, but it's necessary
in order to be able to self-define LIonCore.

PR #26, opened 2023-01-16
Remove 'container' property from NamespacedEntity

dslmeinte:
As agreed last meeting (Friday 2022-01-13). Most impact is in the Programming M3, i.e. the M3's implementation in terms of classes.

### DIFF
--- a/src/m3/types.ts
+++ b/src/m3/types.ts
@@ -4,42 +4,56 @@
  */
 
 import {MultiRef, SingleRef, unresolved} from "../references.ts"
-import {Id, Node} from "../types.ts"
+import {Id, Node as NodeInterface} from "../types.ts"
 import {allFeaturesOf} from "./functions.ts"
 
 
 /**
  * Joins fragments of a qualified name using the `.` character.
  */
-export const qualify = (...names: string[]): string =>
-    names.join(".")
+export const qualify = (...names: (string|undefined)[]): string =>
+    names
+        .filter((name) => typeof name === "string")
+        .join(".")
 
 
-interface NamespaceProvider {
+
+abstract class NodeClass implements NodeInterface {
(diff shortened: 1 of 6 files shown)


src/m3/types.ts
```diff
@@ -4,42 +4,56 @@
...
+export const qualify = (...names: (string|undefined)[]): string =>
+    names
+        .filter((name) => typeof name === "string")
+        .join(".")
 
 
-interface NamespaceProvider {
+
+abstract class NodeClass implements NodeInterface {
+    parent?: NamespaceProvider  // every parent provides a namespace
```
enikao, line 22: (Assuming the context is every model, not only metamodels.)
I don't think every parent is a NamespaceContainer. I might skip a level when creating namespaces, or create two namespaces for one element in the ancestor axis.

dslmeinte: That statement is only correct for M2s (as M3-instances). So, “every parent in an M2 also _happens_ to be a namespace” is more accurate. (Only the following can have children/be parents: `Metamodel`, `Concept`, `ConceptInterface`, `Enumeration`.)

enikao: So `NodeClass` is only used to represent M2 contents?

dslmeinte: Correct. It could therefore receive a better name. Suggestions? It can't be `M3Concept` because that's an already-defined sum type (which is more precise than an unsealed abstract class).

enikao (approved):

src/m3/types.ts
dslmeinte: I'd like to push 1 more commit (with some small improvements) after finding a good name for `NodeClass`.

dslmeinte: Maybe `"M3Node"`? (Or `"M2Node"`?)

PR #27, opened 2023-01-16
Serialize parent's ID as well

dslmeinte:
Per https://github.com/LIonWeb-org/organization/issues/55


enikao:
@dslmeinte Did you see my latest comment in the issue about `children` vs. `containment`? (slightly related to this PR, as it might change the name of the `parent` field).

enikao (approved):

PR #29, opened 2023-01-19
Implement a stdlib of primitive types

dslmeinte:
And replace ad-hoc instantiated with the ones from that stdlib. This includes changes to the M3-specific (de-)serializer.


models/stdlib.json
```diff
@@ -0,0 +1,52 @@
...
+        "LIonCore_stdlib_JSON"
+      ]
+    },
+    "references": {
+      "LIonCore_M3_Metamodel_dependsOn": []
+    }
+  },
+  {
+    "type": "LIonCore_M3_PrimitiveType",
+    "id": "LIonCore_stdlib_String",
```
enikao, line 22: Why are ids of `String` and `JSON` uppercase vs `boolean` and `int` lowercase?

src/m3/functions.ts
```diff
@@ -42,7 +42,7 @@ export const isNonDerivedReference = (feature: Feature): feature is Reference =>
 
 
 /**
- * Determines whether a {@link Feature feature} is “relational“,
+ * Determines whether a {@link Feature feature} is “relational”,
```
enikao, line 45: what's about the double quotes at the end?

dslmeinte: Sorry, my OCD: "this" is different from “this” ;) I've been quite happy since I discovered the keyboard shortcuts for these in macOS...

models/stdlib.json
dslmeinte: According to https://github.com/LIonWeb-org/organization/blob/main/lioncore/metametamodel.adoc#pre-defined-primitive-types it should be `String`, `Boolean`, and `Integer`, indeed. That URL doesn't mention JSON, though: I'm OK with deleting it here, but e.g. the Java impl. does have it.

dslmeinte: Oh, in the interest of meta-circularity: the M3 diagrams should then also use type `Boolean` (for properties), rather than `boolean`, I think.

diagrams/stdlib-gen.puml
```diff
@@ -0,0 +1,27 @@
+@startuml
```
ftomassetti, line 1: Is this file empty?

models/library-imported-from-ecore.json
```diff
@@ -73,7 +51,7 @@
     },
     "references": {
       "LIonCore_M3_Property_type": [
-        "INhBvWyXvxwNsePuX0rdNGB_J9hi85cTb1Q0APXCyJ0"
+        "LIonCore_stdlib_String"
```
ftomassetti, line 54: stdlib makes me think that there are many functionalities provided by this library. What about "common types" or "builtins"?

src/m3/deserializer.ts
```diff
@@ -19,53 +19,61 @@ import {SerializedNode} from "../serialization.ts"
...
+    return map
+}
+
+
 /**
  * Deserializes a metamodel that's serialized into the LIonWeb serialization JSON forma
  * as an instance of the LIonCore/M3 metametamodel, using {@link M3Concept these type definitions}.
  */
-export const deserializeMetamodel = (serializedNodes: SerializedNode[]): Metamodel => {
+export const deserializeMetamodel = (serializedNodes: SerializedNode[], ...dependentMetamodels: Metamodel[]): Metamodel => {
```
ftomassetti, line 38: We would need to discuss if dependencies are transitive or not

src/m3/functions.ts
ftomassetti: For the sake of simplicity I would use the normal quotes, as in `"this"`

src/m3/serializer.ts
```diff
@@ -39,16 +47,13 @@ export const serializeMetamodel = (metamodel: Metamodel): SerializedNode[] => {
...
+                }
             })
             thing.features.forEach(visit)
             return
         }
         if (thing instanceof ConceptInterface) {
-            json.push({
-                type: metaConcepts.conceptInterface.id,
-                id: thing.id,
+            serializeNode(metaConcepts.conceptInterface, {
```
ftomassetti, line 56: I thought we removed the custom serializer. Is that part of a separate PR?

dslmeinte: The generic serializer is still in the works, mostly because of where the DIM2 interacts with PM2: you need to parametrize the generic serializer with a mapping JS object -> M2 element.

diagrams/library-gen.md
```diff
@@ -3,7 +3,7 @@ classDiagram
 
   class Book {
     +String title
-    +int pages
+    +Integer pages
```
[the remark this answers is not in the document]
dslmeinte: Not quite: I wanted to mirror the primitive types we support with their exact names, so with uppercase. Because these diagrams are the result of self-defining the M3, they should use the same names.

diagrams/stdlib-gen.puml
dslmeinte: Shouldn't be: I see it populated.

models/library-imported-from-ecore.json
dslmeinte: You already use "builtins" in the Java impl., right? (Or "builtIns"?)

src/m3/deserializer.ts
dslmeinte: I would say they're _not_ transitive: it's quite easy to just gather all transitive dependencies, which gives you the freedom to determine where you'd get them from, instead of having to provide a loading mechanism (load from disk, from memory, etc.).

diagrams/stdlib-gen.puml
dslmeinte: You mean: there's only comments in there? I can remove these diagrams, as they're not useful, indeed.

src/m3/deserializer.ts
enikao: There's an issue for that: https://github.com/LIonWeb-org/organization/issues/50

models/library-imported-from-ecore.json
ftomassetti: Yes, I currently called it "builtins" but I can definitely change that

src/m3/deserializer.ts
ftomassetti: I think that _not_ transitive is a good idea

diagrams/stdlib-gen.puml
ftomassetti: Yes, sorry, I meant that they seem to contain only comments, so I was not sure if there was something wrong or that was on purpose

dslmeinte: I'll remove them, as they serve no real purpose.

models/library-imported-from-ecore.json
dslmeinte: Name changed.

ftomassetti:
It looks good to me

PR #45, opened 2023-03-04
serializationFormatVersion 1 -> "1"

### LINKED ISSUE #44: Differences in serialization schema
[Here](https://github.com/LIonWeb-org/organization/blob/0d767567e9ccc191d3a787d8c9b19a8a7979e9c2/lioncore/serialization.adoc) the value for `serializationFormatVersion` is indicated as a string, while in the schema in this repository it is indicated as a number.

I think that string can be more general, but I can live with using a number.

I have noticed that while running the integration tests (the Java serialization is not compatible with the JSON schema in this repository)

dslmeinte:
Fixes #44

### DIFF
--- a/models/instance/library.json
+++ b/models/instance/library.json
@@ -1,5 +1,5 @@
 {
-  "serializationFormatVersion": 1,
+  "serializationFormatVersion": "1",
   "nodes": [
     {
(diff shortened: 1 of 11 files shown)


ftomassetti (approved):
+1

Issue #9, opened 2022-12-25
Publish as package for Deno, Node, and the browser


dslmeinte:
Involves:

- Defining what to expose in `src/index.ts`
- Publish as Deno package - see https://deno.land/manual@v1.29.1/advanced/publishing
- Setting up bundling with e.g. https://esbuild.github.io/ for NPM
  - Bundle to `dist/` which also contains NPM metadata?
  - See also https://deno.land/manual@v1.29.1/advanced/publishing/dnt

This issue replaces issues #1 and #5.

alessiostalla:
I've tried using the NPM package in the browser with several shims, but it doesn't seem possible right now. The deno-shim package requires and uses many Node APIs. In particular, now I'm stuck at 
```
mainModule.js:8 Uncaught TypeError: (0 , url_1.pathToFileURL) is not a function
    at ../node_modules/@deno/shim-deno/dist/deno/stable/variables/mainModule.js (mainModule.js:8:1)
    at options.factory (react refresh:6:1)
    at __webpack_require__ (bootstrap:24:1)
    at fn (hot module replacement:62:1)
    at ../node_modules/@deno/shim-deno/dist/deno/stable/variables.js (variables.js:37:1)
    at options.factory (react refresh:6:1)
    at __webpack_require__ (bootstrap:24:1)
    at fn (hot module replacement:62:1)
    at ../node_modules/@deno/shim-deno/dist/deno/stable/functions/chdir.js (chdir.js:10:1)
    at options.factory (react refresh:6:1)
```

b/c the "url" package [doesn't include Node's `pathToFileURL` function](https://github.com/defunctzombie/node-url/issues/63).

The main problem seems to be that Deno's Node shim is monolithic and uses all sorts of Node APIs, which doesn't play nice with the browser.

alessiostalla:
`parse` uses the Deno XML module which in turn uses fs, etc.
I think "dnt" is already taking care of separating the tests from the production code. 
Funny that my spell-checker insists correcting "dnt" to "don't" 🤣

dslmeinte:
Ironic that something (XML parsing) that should be really very portable is an obstacle to portability now...

We can also separate out the parsing-part of the Ecore-importer, keeping just the types that an XML parser is supposed to return, with the remark that the XML parser currently used is compatible with that. What do you think?

alessiostalla:
Yes, we could try that and see if it improves.

dslmeinte:
[...]

PR #46, opened 2023-03-09
Build an NPM package

alessiostalla:
This covers partially issue #9.
I've tested the built package in a Node project. I haven't checked browser compatibility at all.


alessiostalla:
OK, now it works, at least locally. I had to downgrade `nanoid` to the latest 3.x version that, as far as I understand, is maintained alongside 4.x with the same features, but supports CommonJS whereas 4.x only supports ESM (and ESM support on Node is experimental and turned off by default).

deno.lock
```diff
@@ -283,13 +296,18 @@
   "npm": {
     "specifiers": {
       "littoral-templates@0.2.2": "littoral-templates@0.2.2",
+      "nanoid@3.3.4": "nanoid@3.3.4",
```
dslmeinte, line 299: I would expect that version 4.0.0 is really _replaced_ by 3.3.4, but like this it's added. (Also below.)

scripts/build_npm.ts
```diff
@@ -1,6 +1,23 @@
...
+await emptyDir("./npm/script/src/m3/ecore/test");
+await emptyDir("./npm/esm/src/m3/test");
+await emptyDir("./npm/script/src/m3/test");
+
+await copy("diagrams", "npm/esm/diagrams", { overwrite: true });
+await copy("diagrams", "npm/script/diagrams", { overwrite: true });
+await copy("models", "npm/esm/models", { overwrite: true });
+await copy("models", "npm/script/models", { overwrite: true });
+await copy("schemas", "npm/esm/schemas", { overwrite: true });
+await copy("schemas", "npm/script/schemas", { overwrite: true });
```
dslmeinte, line 16: Does this mean that we have now 2 NPM packages: one for ESM and one for CommonJS?

scripts/build_npm.ts
```diff
@@ -1,6 +1,23 @@
...
+await copy("src/m3/ecore/test/library.ecore", "npm/esm/src/m3/ecore/test/library.ecore", { overwrite: true });
```
dslmeinte, line 17: I think all test-artifacts shouldn't end up in `npm/`.

src/m3/test/json-validator.ts
```diff
@@ -25,7 +25,7 @@ export const createJsonValidatorForSchema = (schema: unknown): (json: unknown) =
 }
 
 
-const metaSchema = JSON.parse(Deno.readTextFileSync(`${__dirname(import.meta)}/json.schema.json`))
+const metaSchema = JSON.parse(Deno.readTextFileSync(`src/m3/test/json.schema.json`))
```
dslmeinte, line 28: `__dirname(import.meta)` wasn't working on the Node.js-side?

deno.lock
alessiostalla: Right. I didn't edit the lock file by hand. Maybe there's some other code depending on 4.0.0 that didn't trigger the issue.

scripts/build_npm.ts
alessiostalla: Yes. The CommonJS one is for Node and the ESM one is, I presume, for the browser (but I didn't test it in the browser and I would be surprised if it worked as-is). That's how [Deno's dnt tool](https://github.com/denoland/dnt) works.

scripts/build_npm.ts
alessiostalla: I would think so, too. But I don't know Deno's conventions, and seeing that they are part of the source and the "dnt" tool explicitly runs them and finds them where they are, I assumed that's standard practice for Deno packages. I think we can safely remove them after the build step.

deno.lock
dslmeinte: Running `deno task lock-deps` on the commandline should fix that.

src/m3/test/json-validator.ts
alessiostalla: Nope. It errors with
```
Running tests in ./script/m3/test/library.test.js...

/Users/alessio/projects/lioncore-typescript/npm/script/m3/test/json-validator.js:50
const metaSchema = JSON.parse(dntShim.Deno.readTextFileSync(`${(0, local_files_js_1.__dirname)(import.meta)}/json.schema.json`));
                                                                                                      ^^^^

SyntaxError: Cannot use 'import.meta' outside a module
    at Object.compileFunction (node:vm:352:18)
    at wrapSafe (node:internal/modules/cjs/loader:1031:15)
    at Module._compile (node:internal/modules/cjs/loader:1065:27)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1153:10)
```

etc. etc.

dslmeinte (approved):
Let's merge as-is, and tweak later.

scripts/build_npm.ts
dslmeinte: OK, we'll look how it works exactly after merging ;)

ftomassetti:
Hurray!

PR #52, opened 2023-03-16
NPM package: export more symbols 

alessiostalla:
In particular, JSON serialization/deserialization. Verified on Node and in the browser.


dslmeinte:
@alessiostalla Can you merge from the remote `main` branch? That should take care of some diffs.

src/mod.ts
```diff
@@ -0,0 +1,7 @@
+export * from './types.ts';
```
dslmeinte, line 1: With this the test files should be able to `import { ... }` everything from `../src/mod.ts`, instead of having to “shop around” the entire `src/` directory.

alessiostalla:
> @alessiostalla Can you merge from the remote `main` branch? That should take care of some diffs.

I did, but I hadn't noticed the merge hadn't gone completely well. Now I should have fixed the discrepancies.

src/mod.ts
alessiostalla: Yep! I don't know if `mod.ts` is complete, I suppose not... it's the bare minimum for testing out LionCore in Tylasu, including [serialization/deserialization](https://github.com/Strumenta/tylasu/blob/63b3e3e7664c199227c34e36d79513e674f99188/tests/lionweb.test.ts#L84).

dslmeinte: I think we should add “everything”/as much as possible to `mod.ts` for this to make sense.

dslmeinte:
Sorry, forgot to merge this: now done.

PR #55, opened 2023-08-22
Update for July decisions

dslmeinte:
Main thing is that serialization of LIonCore self-definition and builtins (stdlib) should now match the Java one.

(no Annotations yet)


ftomassetti:
I am not skilled enough with TS to provide a proper review, but I could not see anything immediately wrong

PR #60, opened 2023-08-31
Fix enums

### LINKED ISSUE #57: Enum properties cannot be decoded
Only builtins are decoded, enum properties throw an error.

dslmeinte:
Should fix #57


dslmeinte:
Given that @joswarmer is on holiday, I'll go ahead and merge this.

PR #64, opened 2023-09-12
Migrate from Deno to Node.js

dslmeinte:
Staying close to JS mainstream still has its advantages...


woppa684 (approved):
Just 2 suggestions, the rest looks good to me.

woppa684 (approved):

PR #70, opened 2023-09-21
Update serialization format version

dslmeinte:


src-pkg/serialization.ts
```diff
@@ -48,6 +48,12 @@ export type SerializedLanguageReference = {
     version: string
 }
 
+
+/**
+ * The <em>current</em> version of the serialization format.
+ */
+export const serializationFormatVersion = "2023.1"
```
ftomassetti, line 55: Maybe I would call it "currentSerializationFormatVersion"

ftomassetti (approved):
Looks good

models/instances/library.json
```diff
@@ -1,5 +1,5 @@
 {
-  "serializationFormatVersion": "1",
```
enikao, line 2: I wouldn't do this yet -- there might be more changes.
I already know of one (meta)-change: This field (`serializationFormatVersion`) is currently spec'd as "integer" but will change to "string"

dslmeinte: In the TS and Java implementations this is already a string. I guess we're slightly ahead of the updating of the specification?

Also: these files are not leading in any way.

PR #73, opened 2023-09-30
Integrate lionweb testrunner

joswarmer:


packages/validation/README.md
```diff
@@ -0,0 +1,31 @@
+# @lionweb/validation
+
+The code in this package implements the checks to run the tests in `lionweb-integration-testing/testset`.
+
+This package is intentionally kept very lightweigt, e.g. it uses type instead of classes and has has zero dependencies on external libraries/packages.
+
+It assumes that node.js 18 is used, but will propably work with other versions as well.
+
+To run the tests checkout the `lionweb-integration-testing` project in next to the `lionore-typescript` project
```
dslmeinte, line 9: `lionweb-typescript` in the meanwhile

packages/validation/README.md
```diff
@@ -0,0 +1,31 @@
...
+and run
+
+    # install ts-node
+    yarn
```
dslmeinte, line 13: Can we use NPM here as well? `yarn` would obviously work as well, but it would mean listing another tooling dependency.

packages/validation/build.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env sh
+
+rm -rf dist/ lionweb-core-*.tgz
```
dslmeinte, line 3: Last term should be "`lionweb-validation-*.tgz`".

packages/validation/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "@lionweb/validation",
+  "version": "0.5.0-beta.1",
+  "description": "LionWeb json validation for {Java|Type}Script",
```
dslmeinte, line 4: Twice a "`description`" field. I think I like the second one best.

packages/validation/src/diff/LionwebDiff.ts
```diff
@@ -0,0 +1,215 @@
...
+
+export class LwDiff {
+    errors: string[] = [];
+    lwChecker = new LionWebSyntaxValidator(new ValidationResult());
+
+    constructor() {
+        this.lwChecker.recursive = false;
+    }
+
+    // lwDiff(obj1: any, obj2: any): void {
```
dslmeinte, line 43: Do you have specific plans w.r.t. this commented-out code? If not, I'd prefer to remove it.

dslmeinte:
The last commit on this branch is identical to the last commit on `main`: is that a rebase problem? You should be able to rebase from `main` starting at that commit on `main`. It'd make the diff less confusing. For the end result, it doesn't matter as long as we squash and merge.

dslmeinte (changes requested):

packages/validation/README.md
joswarmer: I'll test whether npm works as well and change this if it does.

packages/validation/build.sh
joswarmer: Is this txz files created somewhere? I don't see it appear when I build the project.

packages/validation/package.json
joswarmer: Fixed

packages/validation/src/diff/LionwebDiff.ts
joswarmer: I had, but not anymore, removed.

packages/validation/README.md
joswarmer: Fixed

packages/validation/README.md
joswarmer: changed all yarn commands to npm, work as well.

packages/validation/build.sh
dslmeinte: They're created by `npm pack`, so it's not relevant for the normal flow, but I like to keep cleaning the repo and the contents of `.npmignore` somewhat in sync.

dslmeinte (approved):

Issue #75, opened 2023-10-10
Propagate existence of annotations


dslmeinte:
See also https://github.com/LionWeb-org/organization/issues/184
This also means e.g. that lots of occurrences of the `Concept` type should be replaced with the `Classifier` type, as well as `conceptOf` &rarr; `classifierOf`. (But isn't a `Classifier` a broader thing than a `Concept`? Are all `Classifier` instances instantiable?)

enikao:
> But isn't a `Classifier` a broader thing than a `Concept`? Are all `Classifier` instances instantiable?

No, but that was also not the case before (abstract concepts).

dslmeinte:
Done on `main`line, per commit [1ab76d8](https://github.com/LionWeb-io/lionweb-typescript/commit/1ab76d83c5a2617eed63297a245662bcc824c189).

PR #77, opened 2023-10-10
Update concept -> classifier including type widening

dslmeinte:
+ (also fix local package deps problem by pointing to fixing versions?)


dslmeinte:
Implements part of issue #75

joswarmer (approved):
This is a lot of changes, so I did not look in detail at each of them.  However it seems quite correct.

PR #83, opened 2023-10-12
Fixed some scripts issues with EACCES

indamutsa:
This PR updated the readme by indicating the user location when running the scripts and adding sudo on some scripts where needed to avoid EACCES -- DELETE BRANCH AFTER MERGE.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,8 @@ Run the following command to build each of the packages:
 
 ```shell
 # Build the project
-./build-all.sh
+# Location: at the root of the project
+sudo ./build-all.sh
 ```
 
 This includes installing any NPM (dev) dependencies.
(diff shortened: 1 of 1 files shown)


joswarmer:
Why would a build and test need sudo?

dslmeinte:
> Why would a build and test need sudo?

I agree that that shouldn't be necessary. In the GH Action log, I don't see an `EACCESS`, but I do see the problem with the non-presence of the `lionweb-integration-testing` repo still, so that can't be (directly) related to root access.

The clarification about in which working directory to execute the scripts: 👍

indamutsa:
I agree. They shouldn't, but when I ran them, they threw EACCES permission denied. So, probably, this can be fixed within the code

dslmeinte:
It seems to work without problems, so no need for the `sudo`s.

PR #85, opened 2023-10-13
Implement a generator that generates TS types from an M2

dslmeinte:
Look at the individual commits! They are nicely separated, so it's not one big diff.


packages/artifacts/chunks/languages/builtins.txt
```diff
@@ -0,0 +1,18 @@
+language LionCore-builtins
+    version: 2023.1
+    entities (↓name):
+
+        primitive type Boolean
+
+        concept-interface INamed
+            features (↓name):
```
enikao, line 8: What is the arrow supposed to mean?

packages/artifacts/chunks/languages/lioncore-types.ts.txt
```diff
@@ -0,0 +1,116 @@
...
+    settings: {
+        multiple?: boolean;
+        annotates?: Classifier;
+        extends?: Annotation;
+        implements: ConceptInterface[];
+    };
+};
+
+/** abstract */ export type Classifier = DynamicNode & {
+    classifier: "Annotation" | "Concept" | "ConceptInterface";
```
enikao, line 26: This "seales" the abstract concept, right? Another language could not define its own concept that extends `Classifier`.

dslmeinte: Correct. I put it in there, because the `shapes` language benefits from it. But I'll only activate that flag for that language then.

packages/artifacts/chunks/languages/builtins.txt
dslmeinte: It's an indication that the features are sorted by name.

packages/artifacts/chunks/languages/builtins-types.ts.txt
```diff
@@ -28,3 +25,4 @@ export type JSON = unknown;
 
 export type String = string;
 
+export type LionCore-builtinsNode = ;
```
enikao, line 28: Is this valid TS? No assigned value?

packages/artifacts/chunks/languages/library-types.ts.txt
```diff
@@ -1,19 +1,16 @@
...
  *     version: 1
  */
 
-
 import {DynamicNode} from "@lionweb/core";
 
-
 export type Book = DynamicNode & {
-    classifier: "Book";
+    // classifier -> Book
```
enikao, line 13: Commented on purpose?

packages/artifacts/chunks/languages/shapes.json
```diff
@@ -8,7 +8,7 @@
   ],
   "nodes": [
     {
-      "id": "/GC1oNIOzuK/OQrQEFJkrCWe88huyTaWVdfRy5Z1zFs=",
+      "id": "degCinkTmS4XeJW0Zt0gH5vpfm/A1DXx/661McUwJYE=",
```
enikao, line 11: Are these ids unstable on purpose?

packages/artifacts/chunks/languages/builtins-types.ts.txt
dslmeinte: Not correct, also because of the invalid ID. Fixed this just now.

packages/artifacts/chunks/languages/library-types.ts.txt
dslmeinte: Yes: `DynamicNode.classifier` has type `Classifier`, so there's no point trying to use discriminated unions here.

packages/artifacts/chunks/languages/shapes.json
dslmeinte: Yes: I changed the name of the language, so the `hashIdGen` 'd IDs changed as well.

PR #89, opened 2023-10-27
Rename SerializedNode.children -> .containments

dslmeinte:
Forgot a couple of places (validations, JSON schema) that need to be changed as well.


joswarmer (approved):
Looks good

PR #90, opened 2023-10-28
Rename own languages

dslmeinte:
This is [issue #195 on specification](https://github.com/LionWeb-io/specification/issues/195).

### DIFF
--- a/packages/core/src/m3/builtins.ts
+++ b/packages/core/src/m3/builtins.ts
@@ -1,18 +1,23 @@
 import {LanguageFactory} from "./factory.js"
 import {lioncoreBuiltinsKey, Property} from "./types.js"
-import {concatenator} from "../utils/string-mapping.js"
+import {StringsMapper} from "../utils/string-mapping.js"
 import {currentReleaseVersion} from "../version.js"
 
 
-const lioncoreBuiltinsIdAndKeyGenerator = concatenator("-")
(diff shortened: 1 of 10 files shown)


packages/core/src/m3/builtins.ts
```diff
@@ -1,18 +1,23 @@
 import {LanguageFactory} from "./factory.js"
 import {lioncoreBuiltinsKey, Property} from "./types.js"
-import {concatenator} from "../utils/string-mapping.js"
+import {StringsMapper} from "../utils/string-mapping.js"
```
enikao, line 3: Should this be here?

enikao (approved):
Only checked for validity, not omissions.

packages/core/src/m3/builtins.ts
dslmeinte: What do you mean? The type of the `lioncoreBuiltinsIdAndKeyGenerator` constant is `StringsMapper`. (I could define that constant without explicitly stating that type, though.)

enikao: I meant: "Should this be in this PR?"
Not really important, it's the only non-pure renaming change.

dslmeinte: Ah yeah, now I get it :) Good question. This particular change comes on the back of cleaning up the whole ID and key generation around LionCore M3, which was a bit too “magical”.

PR #91, opened 2023-10-30
Remove Annotations.multiple

dslmeinte:
See https://github.com/LionWeb-io/specification/issues/32

@joswarmer It seems the validations are not yet complete w.r.t. annotations?


enikao (approved):
looks good, didn't check for omissions

PR #104, opened 2023-12-20
Improve LionWeb-specific JSON diff

joswarmer:
Improved diff, used in lionweb-repository


packages/test/src/validation/diff.test.ts
```diff
@@ -0,0 +1,40 @@
...
+        if (jsonModel === null || jsonModel2 === null) {
+            console.error("Cannot read model files")
+            return
+        }
+        const diff1 = new LionWebJsonDiff()
+        diff1.diffLwChunk(jsonModel, jsonModel)
+        // No errors expected
+        deepEqual(diff1.diffResult.changes, [])
+
+        // const diff = new LionWebJsonDiff()
```
dslmeinte, line 32: Do we want to keep this commented-out code, or can we simply remove it?

packages/validation/src/diff/changes/Change.ts
```diff
@@ -0,0 +1,52 @@
+import { JsonContext } from "../../issues/index.js"
+
+export type ChangeType =
```
dslmeinte, line 3: This starts to feel like a whole separate package. Let's not do that right now, but probably soon-ish, also to aid any ∂ protocol work.

release.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env sh
+
+# release all packages
+
+cd packages
+
+for package in "core" "utilities" "cli" "validation" 
+do
+  (cd $package; npm run release)
```
dslmeinte, line 9: I'm not sure that `npm run release` is clearer than `npm publish` (to which all `release` scripts are equivalent).

packages/validation/package.json
```diff
@@ -19,7 +19,11 @@
     "testInvalid": "node dist/runners/RunCheckFolder.js ../../../lionweb-integration-testing/testset/invalid",
     "testInvalidWithLanguage": "node dist/runners/RunCheckFolderWithLanguage.js ../../../lionweb-integration-testing/testset/withLanguage/invalid/ ../../../lionweb-integration-testing/testset/withLanguage/myLang.language.json",
     "testValidWithLanguage": "node dist/runners/RunCheckFolderWithLanguage.js ../../../lionweb-integration-testing/testset/withLanguage/valid/ ../../../lionweb-integration-testing/testset/withLanguage/myLang.language.json",
-    "test": "mocha"
+    "test": "mocha",
+    "release": "npm publish",
+    "publish-local": "npm publish --registry http://localhost:4873",
```
dslmeinte, line 24: What are these “local” scripts for? If we expose it on `main` then we should at least document it.

packages/core/src/serialization.ts
```diff
@@ -35,11 +35,13 @@ export type SerializedReference = {
  * Type definition for an AST node serialized to JSON.
  */
 export type SerializedNode = {
+    // TODO should be Id
     id: string
     classifier: MetaPointer
     properties: SerializedProperty[]
     containments: SerializedContainment[]
     references: SerializedReference[]
+    // TODO inconsistent with the use of Ids
```
dslmeinte, line 44: You mean: for consistency, we should fix the 1st TODO, and then both are?

dslmeinte (approved):

packages/test/src/validation/diff.test.ts
joswarmer: Can be removed.

packages/validation/src/diff/changes/Change.ts
joswarmer: Yes, I wat thinking the same.

release.sh
joswarmer: Think you are right.

packages/validation/package.json
joswarmer: I use them to do a local `npm publish` with a local npm registry (in my case `verdaccio`). That way I can use an unpublished version in other projects, which is sometimes very handy. We should document this indeed.

packages/core/src/serialization.ts
joswarmer: Inconsistency is between using `Id[]` and `Ids` and `string`. I personally would stick with `Id` and `Id[]`, we use `[]` for other array types (like `SerializedProperty `, etc. ) as well. `Ids` is just obfuscating things.  
So we should change the `string` in the first TODO into `Id` and replace `Ids` (line 21) with `Id[]`.

release.sh
joswarmer: changed

packages/core/src/serialization.ts
joswarmer: fixed

packages/validation/package.json
joswarmer: documented

packages/validation/src/diff/changes/Change.ts
joswarmer: Later ...

packages/test/src/validation/diff.test.ts
joswarmer: Removed

joswarmer (commented):
All done

PR #111, opened 2024-01-16
Diff finds changes in order of annotations, children and targets

joswarmer:
@dslmeinte: this is version 0.6.0-beta.5 of the validation package as needed by lionweb-repository.  Will publish this version as soon as it is merged into main.

### DIFF
--- a/packages/validation/src/json/NodeUtils.ts
+++ b/packages/validation/src/json/NodeUtils.ts
@@ -1,6 +1,7 @@
 import {
+    isEqualMetaPointer,
     isEqualReferenceTarget,
-    LionWebJsonContainment,
+    LionWebJsonContainment, LionWebJsonMetaPointer,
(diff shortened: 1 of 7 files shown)


packages/validation/src/json/NodeUtils.ts
```diff
@@ -1,6 +1,7 @@
 import {
+    isEqualMetaPointer,
     isEqualReferenceTarget,
-    LionWebJsonContainment,
+    LionWebJsonContainment, LionWebJsonMetaPointer,
```
dslmeinte, line 4: Doesn't the linter trigger on this?

dslmeinte (approved):

packages/validation/src/json/NodeUtils.ts
joswarmer: No, it doesn't.

PR #116, opened 2024-01-22
Rework extract functionality

dslmeinte:
* remove shortening
  * re-decompose into 'sort' + 'textualize'
  * extract -> sort
  * some cosmetics w.r.t. cli/validation


packages/cli/README.md
```diff
@@ -15,47 +15,50 @@ $ npx @lionweb/cli
...
 
-## Extracting essential information from a serialization chunk
+This CLI utility does not perform any explicit validation apart from the file at the given path existing and being valid JSON.
+It does some implicit validation as it can error out on incorrect serializations.
+
+
+## Sorting a serialization chunk
 
-Run the following command to make "extractions" from a serialization chunk (e.g.):
+Run the following command to sort a serialization chunk (e.g.):
```
joswarmer, line 24: Why sort at all?

packages/cli/README.md
```diff
@@ -15,47 +15,50 @@ $ npx @lionweb/cli
...
+* Add a `validation` commmand to validate JSON files as serialization chunks.
+* Remove the shortening functionality from `extract`.
+* Rename the `extract` command &rarr; `sort`, and remove the textualization for chunks that are the serialization of languages.
+* Textualizing a serialization chunk of languages will use the LionCore/M3-specific syntax _unless_ the flag `--languagesAsRegular` is an argument.
 
 ### 0.6.0
 
--   Add a `repair` command that aligns a JSON serialization chunk with the specification.
--   Add a `textualize` command – that's optionally language-aware – to render a JSON serialization chunk as pure text.
+* Add a `repair` command that aligns a JSON serialization chunk with the specification.
```
joswarmer, line 51: Can you elaborate what "aligns" means?

packages/cli/src/lionweb-cli.ts
```diff
@@ -69,7 +68,7 @@ If the chunk is the serialization of a LionCore Language/M2, then a textual rend
 `The ${GENERATE_TS_TYPES_COMMAND} command generates a TypeScript source files with type definitions for the given (JSON serializations of) languages, assuming the use of the dynamic façade.`
                 )
             } else {
-                generateTsTypesWith(commandArgs)
+                await generateTsTypesWith(commandArgs)
```
joswarmer, line 71: Why made this async?

packages/cli/README.md
dslmeinte: Mainly for the purpose of being able to compare two chunks for equality. We probably should pull the differ a bit to the fore, instead.

packages/cli/src/lionweb-cli.ts
dslmeinte: `generateTsTypesWith` is already `async`, and `async` is viral :(

packages/cli/README.md
dslmeinte: Align in the sense of matching exactly with the order of the key-value pairs in the specification. This is also in the interest of introspectivity, predictability, etc.

PR #133, opened 2024-02-24
Ensure LionCore-builtins is mentioned in used languages when used

dslmeinte:
(`LionCore-builtins` is not an implicitly-used language.)


joswarmer (approved):
Prima

Issue #137, opened 2024-03-04
Infer (partial) language definition from chunk


dslmeinte:
Infer `Language`s from a given `SerializationChunk`. Obviously, the result is partial (no inheritance data etc.), but could serve as a starting point.

joswarmer:
I would rename this issue to something like "Infer (partial) language definition from chunk. Reconstitute is much more that this.

dslmeinte:
Implemented by @ahmed-negm in PR #141

PR #141, opened 2024-03-08
#137 Infer (partial) language(s) definition from a given serialization chunk

ahmed-negm:


packages/utilities/src/inferLanguages.ts
```diff
@@ -0,0 +1,122 @@
...
+        if (language.entities.filter(entity => entity.key === entityName).length) {
+            continue
+        }
+
+        const concept = new Concept(language, entityName, key(language.name, entityName), id(language.name, entityName), false)
+        language.havingEntities(concept)
+        concepts.set(node.id, concept)
+
+        for (const property of node.properties) {
+            const propertyName = property.property.key.split("_")[2]
```
dslmeinte, line 46: Does this work in general? It assumes that keys have the form `<0>_<1>_<2>_...` and that the <2> part is the only meaningful part, right?

packages/utilities/src/inferLanguages.ts
```diff
@@ -0,0 +1,122 @@
...
+        }
+
+        for (const containment of node.containments) {
+            const containmentName = containment.containment.key
+
+            if (concept.features.filter(feature => feature.key === containmentName).length) {
+                continue
+            }
+
+            // TODO: Check if the containment is multiple or single
```
dslmeinte, line 71: I'd say: if any of the containments encountered is an array with >1 elements, it's multiple – otherwise not.

packages/utilities/src/inferLanguages.ts
```diff
@@ -0,0 +1,122 @@
...
+            const feature = new Containment(
+                concept,
+                containmentName,
+                key(languageName, concept.name, containmentName),
+                id(languageName, concept.name, containmentName)
+            ).isMultiple()
+            concept.havingFeatures(feature)
+
+            // TODO: What if `children` list is empty?
```
dslmeinte, line 80: Should never be empty according to spec. But you could also use the `asArray` function from `@lionweb/core`.

dslmeinte:
Nice and good work, @ahmed-negm ! جيد جداً!

Since this a draft PR I left a couple of single comments. We might also want to add some tests, particularly because some of the logic is non-trivial.

packages/utilities/src/inferLanguages.ts
ahmed-negm: That's how I interpret this key `library_Library_name`. The first part is the language key, the second part is the classifier key, and the third part is the property key. Do I miss something?

packages/utilities/src/inferLanguages.ts
```diff
@@ -0,0 +1,122 @@
...
+        const languageName = node.classifier.language
+        const entityName = node.classifier.key
+
+        const language = getLanguage(languages, languageName)
+        if (language.entities.filter(entity => entity.key === entityName).length) {
+            continue
+        }
+
+        const concept = new Concept(language, entityName, key(language.name, entityName), id(language.name, entityName), false)
+        language.havingEntities(concept)
```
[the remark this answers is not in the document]
ahmed-negm: Could you please explain this further? I have already lookup for languages and retrieve the corresponding concept's language from `node.classifier`:
```
const languageName = node.classifier.language
const language = getLanguage(languages, languageName)
```

ahmed-negm:
Dank Je wel @dslmeinte! Unit tests are in-progress 🙂

packages/utilities/src/inferLanguages.ts
dslmeinte: Any language can choose how its keys itself. The pattern is true for some of our example languages (like library), but not even for the LionCore M3 and built-ins languages (which use "`-`" as the separator). So, in general, this doesn't work.

OTOH: it's the only thing you can do to come up with some kind of meaningful name. Maybe you could extend this a bit with trying to split on various kinds of separators ("`-`", "`_`"), and seeing whether that split matches some kind of notion of a qualified name? But that already suggests a lot of heuristics, so maybe we should stick to something simple/easy that works 70% of the time.

packages/utilities/src/inferLanguages.ts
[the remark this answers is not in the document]
dslmeinte: Ah yes, you're already doing that. I was confused by the identifier `entityName` which is actually the classifier's key.

ahmed-negm:
@dslmeinte could you please have another look?

packages/test/src/utils/inferLanguages.test.ts
```diff
@@ -0,0 +1,53 @@
...
+import {LanguageEntity, serializeLanguages, serializeNodes } from "@lionweb/core"
+import { inferLanguagesFromChunk } from "@lionweb/utilities"
+
+import { libraryExtractionFacade, libraryModel } from "../instances/library.js"
+import { minimalLibraryLanguage } from "../languages/minimal-library.js"
+import { multiExtractionFacade, multiModel } from "../instances/multi.js"
+import { multiLanguage } from "../languages/multi.js"
+
+describe("inferLanguagesFromChunk", () => {
+    it("should correctly infer the minimal library language from the instance", () => {
```
dslmeinte, line 11: Alternatively, we make sure that the example library instance covers the entire library language. That way we wouldn't need to have a minimal library language.

packages/utilities/src/inferLanguages.ts
```diff
@@ -128,6 +130,17 @@ function getLanguage(languages: Map<string, Language>, languageName: string) {
     return language
 }
 
+function getPropertyName(key: string) {
```
dslmeinte, line 133: Getting better, although this heuristic probably fails quite easily, e.g. if there are less than 2 separators. Would it be an idea to write some unit tests for the `getPropertyName` function (which I'd name `deriveLikelyPropertyName` or something like that to indicate it's an educated guess)?

dslmeinte (approved):
But have a look at the comments as well, من فضلك.

packages/test/src/utils/inferLanguages.test.ts
ahmed-negm: I don't think we can infer the entire library language since it involves inheritance which probably cannot be deduced from the instance

dslmeinte: Ah, true! :)

dslmeinte:
btw @ahmed-negm : When you merge this PR, please do so with “Squash and merge”.

dslmeinte (approved):

Issue #106, opened 2023-12-20
Compare textual syntaxes


dslmeinte:
Currently, two textual syntaxes (including their implemented renderers) exist with `lionweb-typescript`: https://github.com/LionWeb-io/lionweb-typescript/blob/main/packages/validation/src/json/LionWebJsonChunkWrapper.ts#L76 and https://github.com/LionWeb-io/lionweb-typescript/blob/feature/generic-textual-syntax/packages/utilities/src/serialization/textual-syntax.ts.
Let's compare these and see whether one suffices. (It might be necessary to make the one in `utilities` language-aware.)

dslmeinte:
This ties in with issue #107. I think it's probably not a bad idea to have a (“low-level”) textual syntax specifically for serialization chunks _only_, and then a separate one that can use M2s to provide a more human readable textual syntax in most cases.

PR #144, opened 2024-03-13
Extend metrics functionality

dslmeinte:
See separate commits for what's changed.


joswarmer (approved):

dslmeinte:
Sorry, first need to add more unit tests before it's good to go.

dslmeinte:
> There is a `textualizer.ts` both in `utilities/src/m3` and a `textualizer.ts` in `utilities/src/serialization` and another one in `cli/src/serialization`. Maybe they should get more distinctive names.

I did some renaming so that `cli/src/serialization/textualizer.ts` is now `textualize-command.ts`. For the other two we have issue #106.

Issue #143, opened 2024-03-12
classBasedClassifierDeducerFor doesn't work with bundlers/minifiers


alessiostalla:
`classBasedClassifierDeducerFor` uses the `constructor.name` property that bundlers and minifiers will potentially change. I managed to break it just by importing an unrelated "Property" function; the bundler decided to rename Lionweb's Property class to Property2, thus making it impossible to deserialize any language.

In Tylasu we solved this with an optional property one can add to classes to fix their name, and a decorator that makes setting the property easier. Something like:

```typescript
@Classifier("Property")
export class Property {
  ...
}
```

dslmeinte:
The easiest way to fix this is probably to add a field `metaType` to the abstract class `M3Node` and populate that with the correct string.

dslmeinte:
See PR #146

dslmeinte:
Fixed by PR #146

PR #146, opened 2024-03-25
Fix/replace use of constructor name

dslmeinte:
Fixes issue #143

@alessiostalla: could you have a look as well?


alessiostalla:
Looks good to me, thanks!

PR #158, opened 2024-05-28
Support deserialization of custom primitive types

### LINKED ISSUE #156: Custom primitive types / built-in types
I see in `deserializeBuiltin` there is no way to deserialize a custom type while Lionweb/Java supports this (https://github.com/LionWeb-io/lionweb-java/blob/8b7bb5d9dee3f9687ea6847c72afb1aea85847b1/core/src/main/java/io/lionweb/lioncore/java/serialization/PrimitiveValuesSerialization.java#L56-L60). 
We now have encountered models serialized with that feature from Java that we cannot deserialize in JS.
Is this planned? Is there a workaround?

ftomassetti:
Based on the discussion we had during the meeting on the 25th of May, my understanding is that the different implementation of the libraries can provide custom deserializers for primitive types, as this is not something defined by the specification (neither required or forbidden).

So I thought I could put together an initial implementation to start the conversation.

In this PR we:
- Permit to define custom primitive type deserializers
- We convert the builtin primitive types to use those primitive type deserializers. That also means that someone could, potentially, override the default way in which primitive types are serialized
- Add a simple test to check the mechanism

If the general approach works I could also add support for primitive type serializer.

I am no TS or functional code expert, so I thought it made sense to open a PR early to check if I am going in a direction that makes sense for the library. Feedback is very well appreciated :)

Fix #156

### DIFF
--- a/packages/core/src/deserializer.ts
+++ b/packages/core/src/deserializer.ts
@@ -4,7 +4,6 @@ import {InstantiationFacade} from "./facade.js"
 import {MemoisingSymbolTable} from "./symbol-table.js"
 import {Classifier, Containment, Enumeration, Language, PrimitiveType, Property, Reference} from "./m3/types.js"
 import {allFeaturesOf} from "./m3/functions.js"
-import { PrimitiveTypeSerializer } from "./m3/builtins.js"
 import {groupBy} from "./utils/map-helpers.js"
 
 
@@ -19,6 +18,9 @@ const byIdMap = <T extends { id: Id }>(ts: T[]): { [id: Id]: T } => {
     return map
 }
 
+export interface PrimitiveTypeDeserializer {
+    deserializeValue(value: string | undefined, property: Property): unknown | undefined
+}
 
 /**
  * @return a deserialization of a {@link SerializationChunk}
@@ -31,7 +33,7 @@ const byIdMap = <T extends { id: Id }>(ts: T[]): { [id: Id]: T } => {
 export const deserializeSerializationChunk = <NT extends Node>(
     serializationChunk: SerializationChunk,
     instantiationFacade: InstantiationFacade<NT>,
-    dataTypeSerializer: PrimitiveTypeSerializer,
+    primitiveTypeDeserializer: PrimitiveTypeDeserializer,
     languages: Language[],
     // TODO  facades <--> languages, so it's weird that it looks split up like this
     dependentNodes: Node[]
@@ -92,7 +94,7 @@ export const deserializeSerializationChunk = <NT extends Node>(
                     if (property.key in serializedPropertiesPerKey) {
                         const value = serializedPropertiesPerKey[property.key][0].value
                         if (property.type instanceof PrimitiveType) {
-                            propertySettings[property.key] = dataTypeSerializer.deserializeValue(value, property as Property);
+                            propertySettings[property.key] = primitiveTypeDeserializer.deserializeValue(value, property as Property);
                             return
                         }
                         if (property.type instanceof Enumeration) {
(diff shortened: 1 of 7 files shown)


dslmeinte (changes requested):

packages/core/src/deserializer.ts
```diff
@@ -31,6 +31,7 @@ const byIdMap = <T extends { id: Id }>(ts: T[]): { [id: Id]: T } => {
 export const deserializeSerializationChunk = <NT extends Node>(
     serializationChunk: SerializationChunk,
     instantiationFacade: InstantiationFacade<NT>,
+    dataTypeSerializer: PrimitiveTypeSerializer,
```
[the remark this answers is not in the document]
ftomassetti: renaming to `primitiveTypeDeserializer` and making an interface for `PrimitiveTypeDeserializer`

packages/core/src/m3/builtins.ts
```diff
@@ -61,6 +61,8 @@ lioncoreBuiltins.havingEntities(
 
 
 type BuiltinPrimitive = string | boolean | number | Record<string, unknown> | Array<unknown>
+type PrimitiveTypeValue = BuiltinPrimitive | any
```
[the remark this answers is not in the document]
ftomassetti: Done

ftomassetti:
@dslmeinte thank you for the review, I should have applied the suggested corrections

packages/core/src/deserializer.ts
```diff
@@ -31,6 +33,7 @@ const byIdMap = <T extends { id: Id }>(ts: T[]): { [id: Id]: T } => {
 export const deserializeSerializationChunk = <NT extends Node>(
     serializationChunk: SerializationChunk,
     instantiationFacade: InstantiationFacade<NT>,
+    primitiveTypeDeserializer: PrimitiveTypeDeserializer,
```
dslmeinte, line 36: Sorry for not realizing this earlier: this would break the API of the package. Is it possible to move this argument to the last position, and have it have a default value?

packages/core/src/m3/builtins.ts
```diff
@@ -95,26 +98,41 @@ const serializeBuiltin = (value: BuiltinPrimitive): string => {
     throw new Error(`can't serialize value of built-in primitive type: ${value}`)
 }
 
+export class SimplePrimitiveTypeDeserializer implements PrimitiveTypeDeserializer {
```
dslmeinte, line 101: Maybe `DefaultPrimitiveTypeDeserializer`? It's not particularly simple.

dslmeinte (changes requested):
Sorry: some more suggested refactorings...

Also: merge using “Squash...”, please!

ftomassetti:
> Sorry: some more suggested refactorings...

Thank you, I think they are helping making the solution better

packages/core/src/deserializer.ts
ftomassetti: Good idea, done

packages/core/src/m3/builtins.ts
ftomassetti: Done!

packages/core/src/deserializer.ts
dslmeinte: Uhm, I'm seeing that argument still in the 3rd position (and without a default value).

dslmeinte (changes requested):

dslmeinte (approved):

ftomassetti:
@dslmeinte sorry for all the troubles. At some point WebStorm did not push the commits and I did not noticed that

PR #163, opened 2024-06-20
Fix cycle in imports in lionweb/core

dslmeinte:
And immediately release core 0.6.7, etc. as well


ftomassetti (approved):
Looks good to me

Issue #164, opened 2024-07-16
Fix deserialization to work with meta-circular language definitions


dslmeinte:

joswarmer:
What is a meta-circular language definition?

dslmeinte:
E.g. a language definition that defines an annotation which is used to annotate the language itself.

PR #167, opened 2024-07-30
relax deserialization

dslmeinte:
Relax deserialization to just skip serialized nodes for which the corresponding M2 data can't be found.

(This effectively addresses issue #164 .)

### DIFF
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -32,6 +32,11 @@ The following is a list of links to potential starting points:
 
 ## Changelog
 
+### 0.6.8
+
+* Relax deserialization to just skip serialized nodes for which the corresponding M2 data can't be found.
+
+
 ### 0.6.7
 
 * Fix cycle in imports that some bundlers were having trouble with.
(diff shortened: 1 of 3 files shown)


joswarmer:
When a node is not being instantiated this means that all its children (recursively) will not be instantiated either, even if their classifier is perfectly ok.  I understand it is unclear what you would do with such children, as they are lacking a parent.

It might be interesting to also return a list of nodes that could not be instantiated, so the caller at least gets the information and is able to use it (or ignore it). The caller can then choose to do a console.log if necessary.

joswarmer (approved):

dslmeinte:
> When a node is not being instantiated this means that all its children (recursively) will not be instantiated either, even if their classifier is perfectly ok. I understand it is unclear what you would do with such children, as they are lacking a parent.
> 
> It might be interesting to also return a list of nodes that could not be instantiated, so the caller at least gets the information and is able to use it (or ignore it). The caller can then choose to do a console.log if necessary.

I want to address these comments/ideas as part of a roadmap towards a version 1.0. Various things need a bit of a reworking anyway.

PR #171, opened 2024-08-02
Fix null property value in seralization

dslmeinte:


packages/utilities/src/m3/infer-languages.ts
```diff
@@ -58,12 +58,16 @@ export const inferLanguagesFromSerializationChunk = (chunk: SerializationChunk):
                 id(languageName, concept.name, propertyName)
             ).havingKey(property.property.key)
 
-            if (isBoolean(property.value)) {
-                feature.ofType(booleanDatatype)
-            } else if (isNumeric(property.value)) {
-                feature.ofType(integerDatatype)
+            if (property.value === null) {
+                feature.isOptional()
```
joswarmer, line 62: Surprised at the naming convention here. Usually an `isSomething()` method returns whether the `Something` is true (or not).  But this is more of a a `setOptional(...)`

joswarmer (approved):

packages/utilities/src/m3/infer-languages.ts
dslmeinte: It's part of the fluent-style API for the TypeScript LionCore M3. I like the style, but I also understand how it can be confusing.

joswarmer: I understand, never found the fluent style very appealing.

PR #172, opened 2024-08-05
Don't let deserialization throw on unresolvable references

dslmeinte:


packages/core/README.md
```diff
@@ -34,10 +34,12 @@ The following is a list of links to potential starting points:
 
 ### 0.6.8
 
-* Relax deserialization to just skip serialized nodes for which the corresponding M2 data can't be found.
+* With respect to deserialization:
+  * Relax deserialization to just skip serialized nodes for which the corresponding M2 data can't be found.
+  * Fix a bug w.r.t. enumeration literals — deserializing changed the keys of enumeration literals in the language's definition.
```
enikao, line 39: Are there tests for the latter two changes?

dslmeinte: (If I had been able to write a failing test for the 2nd item, I would have found and fixed it ages ago! :))

joswarmer (approved):

PR #173, opened 2024-08-09
Discuss (formulate) roadmap for v1

dslmeinte:


ftomassetti (approved):
Not an expert, but it looks good to me. I should take inspiration from this to plan for the release of LW Java 1.0

ROADMAP-v1.md
```diff
@@ -0,0 +1,36 @@
...
+  * `validation`: validation of serialization chunks — as standalone as possible
+  * `utilities`: ?
+  * `queries`: ?
+
+* Rename `InstantiationFacade` &rarr; `Factory`
+  * And `ExtractionFacade` &rarr; `Reflector`?
+
+* Give deserialization a better API:
+  * Pair up languages and their factories
+  * Make deserialization more configurable:
```
joswarmer, line 23: I favor of this, but we should make sure that the usage should remain simple in case you don't need all those configurable options.

ROADMAP-v1.md
```diff
@@ -0,0 +1,36 @@
...
+    * How to deal with unknown classifiers: dynamic instantiation (i.e.: a fall-back factory for the `DynamicNode` type), or ignore
+    * Provide a channel to report back (and possibly even “converse”) about unknown classifiers, in such a way that we can distinguish between annotations and concepts
+  * This is based on the following principles:
+    * We can deal with broken models, so a “small” problem in the serialization should not prevent the deserialization as a whole.
+      (The GPL types that the model is deserialized into might have a different opinion about it, but that's its problem.)
+    * We should uncover _all_ problems during deserialization, not just the first one and then quit.
```
joswarmer, line 29: This is indeed good to have.

ROADMAP-v1.md
```diff
@@ -0,0 +1,36 @@
...
+    * (“Things should be as FP-style as possible, so effectively a `flatMap` of-sorts.)
```
joswarmer, line 30: I don't think we need to strive to any specific style, as long as it is understandable and usable.

joswarmer (approved):

ROADMAP-v1.md
dslmeinte: Added an extra bullet point about that.

ROADMAP-v1.md
dslmeinte: Modified this remark a bit.

PR #183, opened 2024-11-10
Identity builtin primitive types

dslmeinte:
(I'm pretty sure this was reported as an issue, but I can't find by whom anymore.)

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@lionweb/cli",
-  "version": "0.6.11-beta.0",
+  "version": "0.6.11-beta.2",
   "description": "LionWeb CLI for {Java|Type}Script",
   "bin": {
     "lionweb-cli": "./dist/lionweb-cli.js"
@@ -16,8 +16,8 @@
     "url": "https://github.com/LionWeb-io/lionweb-typescript/issues"
   },
   "dependencies": {
-    "@lionweb/core": "0.6.10-beta.0",
-    "@lionweb/utilities": "0.6.10-beta.0",
+    "@lionweb/core": "0.6.10-beta.2",
+    "@lionweb/utilities": "0.6.10-beta.2",
     "@lionweb/validation": "0.6.3"
   },
   "scripts": {
@@ -25,7 +25,10 @@
     "build": "tsc",
     "lint": "eslint src",
     "test": "node dist/lionweb-cli.js",
+    "pre-release-either": "npm run clean && npm install && npm run build",
+    "prerelease": "npm run pre-release-either",
     "release": "npm publish",
+    "prerelease-beta": "npm run pre-release-either",
     "release-beta": "npm publish --tag beta"
   }
 }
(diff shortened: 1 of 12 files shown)


packages/cli/package.json
```diff
@@ -16,16 +16,19 @@
...
+    "@lionweb/core": "0.6.10-beta.2",
+    "@lionweb/utilities": "0.6.10-beta.2",
     "@lionweb/validation": "0.6.3"
   },
   "scripts": {
     "clean": "rm -rf dist/ lionweb-cli-*.tgz",
     "build": "tsc",
     "lint": "eslint src",
     "test": "node dist/lionweb-cli.js",
+    "pre-release-either": "npm run clean && npm install && npm run build",
```
ftomassetti, line 28: what pre-release either means in this case?

packages/core/README.md
```diff
@@ -32,6 +32,12 @@ The following is a list of links to potential starting points:
 
 ## Changelog
 
+### 0.6.10 - not officially released yet
+
+* Made `DefaultPrimitiveTypeDeserializer` and `DefaultPrimitiveTypeSerializer` be able to deal with duplicate definitions of datatypes.
+  * Expose a function `shouldBeIdentical` that determines whether two datatypes should be structurally equal based on equality of: meta type, key, and language's key.
```
ftomassetti, line 38: If I read "structurally equal" I expect that two datatypes with the same fields would be considered equal, while from the rest of the description I think we look at the metapointers and not the structure of the datatype

packages/core/src/m3/builtins.ts
```diff
@@ -83,19 +83,44 @@ const builtinFeatures = {
     inamed_name
 }
 
-export class DefaultPrimitiveTypeDeserializer implements PrimitiveTypeDeserializer {
 
-    private deserializerByType = new Map<Datatype, SpecificPrimitiveTypeDeserializer>()
+/**
+ * Determines whether two datatypes should be structurally equal based on equality of: meta type, key, and language's key.
```
ftomassetti, line 88: Why calling the method "should"? I would expect it to do behave as an assert method and through an exception if the two elements are not equals.
I would find a method name like `areIdentical` more obvious

ftomassetti (approved):
It looks very good to me. My comments are solely about terminology, but they could be due to me not being familiar with TS conventions

packages/cli/package.json
dslmeinte: It's the pre-command for either the `release` or `release-beta` scripts.

packages/core/README.md
dslmeinte: (see below)

packages/core/src/m3/builtins.ts
dslmeinte: Because two instance of a `Datatype` **should** be equal if their metadata (metapointers and metatypes) are identical. I could also compare them really structurally, but then their metadata should also be identical. And I wouldn't know how to interpret two datatypes that have nonidentical metadata but are otherwise structurally the same, in this context.

PR #184, opened 2024-11-11
Fixes regarding serialization with null children, and null resolveInfo

dslmeinte:


ftomassetti (approved):
I may be out of my depth here, but nothing seems immediately wrong to me

PR #201, opened 2025-01-29
Release of core@0.6.12, utilities@0.6.12, cli@0.6.13

dslmeinte:


dslmeinte:
Notice: I found a bug, so this PR is *on hold* (for the moment)

dslmeinte:
> Notice: I found a bug, so this PR is _on hold_ (for the moment)

Never mind: bug was somewhere else :)

packages/cli/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## 0.6.13
+
+* (Depend on `@lionweb/core` and `@lionweb/utilities` version 0.6.12.)
```
ftomassetti, line 5: Why the parenthesis?

My understanding is that we typically list features and this is more of a comment so we distinguish it by using the parenthesis. I would however consider removing the parenthesis and the bullet point. We could also make it in italic.

packages/core/src/m3/factory.ts
```diff
@@ -32,44 +38,60 @@ export class LanguageFactory {
     }
 
 
-    // TODO  also take care of containment immediately? (then most calls to .having<X> would be unnecessary/wrong)
-
     annotation(name: string, extends_?: SingleRef<Annotation>) {
```
ftomassetti, line 41: question due to my ignorance with TS: shouldn't we specify the return type for this and the following methods?

packages/core/src/m3/functions.ts
```diff
@@ -48,6 +48,34 @@ const isContainment = (feature: Feature): feature is Containment =>
...
     feature instanceof Reference
 
+const isMultiple = (feature: Feature): feature is Link =>
+    feature instanceof Link && feature.multiple
+
+
+/**
+ * The (names of the) metatypes of a feature.
+ */
+type FeatureMetaType =
```
ftomassetti, line 58: Not a TS expert here, but would something like this be type-safe? I mean, protecting from typos in these strings?

```
const FeatureMetaMap = {
    Containment,
    Property,
    Reference,
} as const;

type FeatureMetaType = keyof typeof FeatureMetaMap;
```

packages/core/src/serializer.ts
```diff
@@ -20,14 +21,58 @@ export interface PrimitiveTypeSerializer {
...
     extractionFacade: ExtractionFacade<NT>,
-    primitiveTypeSerializer: PrimitiveTypeSerializer = new DefaultPrimitiveTypeSerializer()
+    primitiveTypeSerializerOrOptions?: PrimitiveTypeSerializer | SerializationOptions
 ): SerializationChunk /* <=> JSON */ => {
+    const primitiveTypeSerializer = (
+        isPrimitiveTypeSerializer(primitiveTypeSerializerOrOptions)
+            ? primitiveTypeSerializerOrOptions
+            : primitiveTypeSerializerOrOptions?.primitiveTypeSerializer
+    ) ?? new DefaultPrimitiveTypeSerializer()
+    const serializeEmptyFeatures = isPrimitiveTypeSerializer(primitiveTypeSerializerOrOptions)
```
ftomassetti, line 72: here shouldn't we look at primitiveTypeSerializer and not primitiveTypeSerializerOrOptions?

packages/test/src/infer-languages.test.ts
```diff
@@ -1,73 +1,63 @@
...
         const serializationChunk = serializeNodes(multiModel, multiExtractionFacade)
 
         const languages = inferLanguagesFromSerializationChunk(serializationChunk)
-        deepEqual(languages.length, 2)
+        equal(languages.length, 2)
         const inferredMultiLanguage = languages[0]
         const inferredLibraryLanguage = languages[1]
 
-        // We can't infer the language dependency
+        // (we can't infer the language dependency)
```
ftomassetti, line 35: Why the parenthesis?

packages/test/src/languages/library.ts
```diff
@@ -12,44 +12,28 @@ export const libraryLanguage = factory.language
...
-const book_type = factory.property(book, "type").ofType(bookType).isOptional()
-book.havingFeatures(book_title, book_pages, book_author, book_type)
-
-const writer_name = factory.property(writer, "name").ofType(stringDatatype).havingKey("library_Writer_name")
-writer.havingFeatures(writer_name)
-// Note: writers are _not_ contained in the library (node) itself, so are separate nodes.
 
-const guideBookWriter_countries = factory.property(guideBookWriter, "countries").ofType(stringDatatype)
-guideBookWriter.havingFeatures(guideBookWriter_countries)
+factory.property(library, "name").ofType(stringDatatype).havingKey("library_Library_name")
```
ftomassetti, line 25: much cleaner!

packages/utilities/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## 0.6.12
+
+* (Depend on `@lionweb/core` version 0.6.12.)
```
ftomassetti, line 5: Same as above

ftomassetti (approved):
I took a look and to me nothing seems immediately wrong

packages/cli/CHANGELOG.md
dslmeinte: Good point! We could list it as a separate, non-list characteristic of each release: just mentioning the relevant (and in most cases: only) upstream dependencies.

packages/core/src/m3/factory.ts
dslmeinte: Actually: we should. Didn't do that so far because the API was involving a bit, still.

packages/core/src/m3/functions.ts
dslmeinte: That's actually (effectively) the same thing. Using the sum type already prevents typos: a typo would prevent compilation.

packages/core/src/m3/factory.ts
dslmeinte: Oh, and specifying a return type is optional in TS as TS derives it from the statement block, but one can certainly consider it to be good manners to specify it nevertheless.

packages/test/src/infer-languages.test.ts
dslmeinte: Good point: clarified the comment.

packages/core/src/serializer.ts
dslmeinte: Line 68 checks whether `primitiveTypeSerializerOrOptions` *is* a `PrimitiveTypeSerializer`.

PR #209, opened 2025-04-15
`tryFromId` in `IdMapping`

dslmeinte:


packages/class-core/src/id-mapping.ts
```diff
@@ -42,11 +42,15 @@ export class IdMapping {
         return this.nodesById[id];
     }
 
-    fromRefId(idOrUnresolved: IdOrUnresolved): SingleRef<INodeBase> {
-        return idOrUnresolved === null
+    tryFromId = (id: Id): (INodeBase | undefined) =>
+        id in this.nodesById
```
joswarmer, line 46: You might want to use a Map instead of an object type.
Then you can simply write this as

```
type NodesById: Map<Id, INodeBase>

   tryFromId = (id: Id): (INodeBase | undefined) =>
          this.nodeById.get(id)
```

I find using a Map for this is conceptually more clean.
Also, this does the lookup once instead of twice so it is more efficient.

joswarmer (approved):

packages/class-core/src/id-mapping.ts
dslmeinte: Actually: `this.nodeById[id]` also already achieves that, so I'll change it to that and stick to using an object for now. I'll add a TODO about using a `Map` instead, though, OK?

PR #218, opened 2025-06-27
Fix/windows build

### LINKED ISSUE #217: Build support on Windows
The current build process does not work on Windows.
I will report here the minor issues I found and will provide PRs to fix them.

1. `source` is not available in Windows
2. `mkdir -p dir`has problems on Windows. Alternative: `npx mkdirp metrics` via https://stackoverflow.com/questions/39727920/cant-use-mkdir-in-npm-script-from-windows

pjmolina:
# Fixes #217 Build support for Windows

1. avoid using `source`
2. adds shebang for bash files
3. uses `npx mkdirp` instead `mkdir` for better portability
4. execute with shell `sh ./make.sh`
5. fix broken tests on Windows due to EOL differences
6. fix **.gitignore** to ignore descedant node_module folders also
7. update versions for `types/node` and `prettier`
8. **package.json** automatically reformatted accordingly to `.editorconfig` rules (4 spaces)

Now the lib builds in Windows also.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,5 +3,4 @@
 /lionweb-typescript.iml
 
 # NPM:
-/node_modules
-
+node_modules/
--- a/package.json
+++ b/package.json
@@ -1,52 +1,52 @@
 {
-  "name": "lionweb-typescript",
-  "version": "0.7.0-dev",
-  "description": "LionWeb core for {Java|Type}Script",
-  "license": "Apache-2.0",
-  "repository": {
-    "type": "git",
-    "url": "git+https://github.com/LionWeb-io/lionweb-typescript.git"
-  },
-  "bugs": {
-    "url": "https://github.com/LionWeb-io/lionweb-typescript/issues"
-  },
-  "devDependencies": {
-    "@types/node": "22.15.21",
-    "@typescript-eslint/eslint-plugin": "6.21.0",
-    "@typescript-eslint/parser": "6.21.0",
-    "prettier": "3.6.2",
-    "typescript": "5.3.3"
-  },
-  "scripts": {
-    "initialize": "npm run clean && npm install && npm run setup && npm run build && sh ./make.sh",
-    "clean": "rm -rf node_modules/ && npm run clean -w packages --if-present",
-    "build": "npm run build -w packages --if-present",
-    "lint": "npm run lint -w packages --if-present",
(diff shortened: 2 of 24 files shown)


.gitignore
```diff
@@ -3,5 +3,4 @@
 /lionweb-typescript.iml
 
 # NPM:
-/node_modules
-
+node_modules/
```
dslmeinte, line 6: This doesn't change the semantics from “just `node_modules` at root level” to “any `node_modules` directory”?

package.json
```diff
@@ -1,51 +1,51 @@
...
+    "license": "Apache-2.0",
+    "repository": {
+        "type": "git",
+        "url": "git+https://github.com/LionWeb-io/lionweb-typescript.git"
+    },
+    "bugs": {
+        "url": "https://github.com/LionWeb-io/lionweb-typescript/issues"
+    },
+    "devDependencies": {
+        "@types/node": "24.0.4",
```
dslmeinte, line 14: These versions should be determined by the `versions.json` file, and are updated through running the `update-package-versions.js` script: see [this section of the README](https://github.com/LionWeb-io/lionweb-typescript?tab=readme-ov-file#version-numbers).

dslmeinte (changes requested):

.gitignore
dslmeinte: `node_modules/` is fine, actually. Some packages will anyway have their own `node_modules/`.

package.json
```diff
@@ -14,11 +14,11 @@
     "@types/node": "22.15.21",
     "@typescript-eslint/eslint-plugin": "6.21.0",
     "@typescript-eslint/parser": "6.21.0",
-    "prettier": "3.5.3",
+    "prettier": "3.6.2",
     "typescript": "5.3.3"
   },
   "scripts": {
-    "initialize": "npm run clean && npm install && npm run setup && npm run build && source make.sh",
+    "initialize": "npm run clean && npm install && npm run setup && npm run build && sh ./make.sh",
```
dslmeinte, line 21: Let's also update the README to use the shebang formulation.

dslmeinte:
Can we also replace any `rm -rf` with `[npx] rimraf`?

package.json
pjmolina: Done

PR #220, opened 2025-07-02
Homogeneus indent to 4 spaces

pjmolina:
Set homogeneus indent to 4 spaces.

There are 27 places where `JSON.stringify(` is used. 

@dslmeinte Feel free to decide/review which one needs to use 4 spaces or nor.

### DIFF
--- a/packages/ts-utils/src/index.ts
+++ b/packages/ts-utils/src/index.ts
@@ -1,6 +1,7 @@
 export * from "./array-helpers.js"
 export * from "./cycles.js"
 export * from "./graphs.js"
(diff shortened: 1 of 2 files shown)


dslmeinte:
> There are 27 places where `JSON.stringify(` is used.
> @dslmeinte Feel free to decide/review which one needs to use 4 spaces or nor.

I think we should have central aliases for `JSON.stringify(<json>)` and `JSON.stringify(<json>, null, 4)` in the `ts-utils` package: what do you think?

pjmolina:
Yes. Makes sense to me.

PR #226, opened 2025-07-18
homogeneus indent to 4 spaces

pjmolina:
homogeneus indent to 4 spaces

### DIFF
--- a/packages/ts-utils/src/json.ts
+++ b/packages/ts-utils/src/json.ts
@@ -1,0 +1,8 @@
+/** Normalized JSON with 4 spaces as indent */
+export const asPrettyJsonString = (obj: unknown): string => {
+    return JSON.stringify(obj, null, 4);
+}
+/** Minimal JSON with no spaces as indent */
+export const asMinimalJsonString  = (obj: unknown): string => {
+    return JSON.stringify(obj, null, 0);
+}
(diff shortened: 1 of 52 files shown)


packages/ts-utils/src/json.ts
```diff
@@ -0,0 +1,4 @@
+/** Normalized JSON with 4 spaces a as indent */
```
dslmeinte, line 1: (superfluous "a")

packages/ts-utils/src/json.ts
```diff
@@ -0,0 +1,4 @@
...
+export const toJson = (obj: unknown): string => {
```
dslmeinte, line 2: I'd rather call this `asIndentedJsonString` (or `asPrettyJsonString`) because `obj` already _is_ JSON, or you couldn't call `JSON.stringify` with it anyway.

In addition, I'd like to implement a function `asMinimalJsonString` that just does `JSON.stringify`.

Then, we should replace all `JSON.stringify(<obj>)` with `asMinimalJsonString`, and all other `JSON.stringify`s with `asPrettyJsonString`.

dslmeinte (changes requested):

dslmeinte:
@pjmolina Can you have a look whether you agree with my additional commits?

pjmolina:
Looks good to me. 
Ready to merge on my side.

PR #227, opened 2025-07-21
Implemented queues in LionWeb delta protocol client

dslmeinte:


packages/deltas-websocket/src/utils/priority-queue.ts
```diff
@@ -0,0 +1,94 @@
...
+ */
+export type PriorityQueueAcceptor<T> = (t: T) => void
+
+/**
+ * @return a {@link PriorityQueueAcceptor} function that accepts instances of `T`,
+ * passing them to the `process` function in (min-max) priority order,
+ * as computed by the `priorityFunc`, starting with the `firstPriority`.
+ * @throws an {@link Error error} when a `T` is passed with a priority that has already been processed.
+ */
+export const priorityQueueAcceptor = <T>(priorityFunc: (t: T) => number, firstPriority: number, process: Procedure<T>): PriorityQueueAcceptor<T> => {
```
enikao, line 61: Nice design to factor out this functionality

enikao (commented):
looks ok to me logically, can't really judge the TS implementation.

packages/deltas-websocket/src/utils/priority-queue.ts
dslmeinte: Also makes it a lot easier – read: *possible* – to unit test the tricky behavior of the queue!

PR #228, opened 2025-07-23
Implement queries

dslmeinte:


packages/deltas-websocket/src/client/client-impl.ts
```diff
@@ -168,33 +167,82 @@ export class LionWebClient {
...
         })
 
-    signOn = async (queryId: LionWebId) =>
+    async subscribeToChangingPartitions(queryId: LionWebId, parameters: SubscribeToPartitionChangesParameters): Promise<void> {
         await this.enqueueQuery({
+            messageKind: "SubscribeToChangingPartitionsRequest",
+            queryId,
+            ...parameters,
+            protocolMessages: []
+        } as SubscribeToChangingPartitionsRequest)
```
enikao, line 187: shouldn't this be `...Response`?

packages/deltas-websocket/src/client/client-impl.ts
```diff
@@ -168,33 +167,82 @@ export class LionWebClient {
...
 
-    async disconnect() {
+    async reconnect(queryId: LionWebId, participationId: LionWebId, lastReceivedSequenceNumber: number): Promise<void> {
+        const response = await this.enqueueQuery({
+            messageKind: "ReconnectRequest",
+            queryId,
+            participationId,
+            lastReceivedSequenceNumber,
+            protocolMessages: []
+        } as ReconnectRequest) as ReconnectResponse
```
enikao, line 240: depending how much error handling we're already considering, this might be an error event telling the participation is outdated.

packages/deltas-websocket/src/payload/query-types.ts
```diff
@@ -15,47 +15,119 @@
...
+export interface QueryMessage extends DeltaProtocolMessage {
     queryId: LionWebId
 }
 
-export interface QueryResponse extends DeltaProtocolMessage {
-    queryId: LionWebId
+
+// in order of the specification:
+
+export interface SubscribeToPartitionChangesParameters {
```
enikao, line 32: this doesn't exist in the spec, will it be flattened in the serialization?

packages/deltas-websocket/src/client/client-impl.ts
dslmeinte: No, this is the request part: it gets en-Q'd, meaning that the message is sent, and then the `resolve` callback of the associated `Promise` is stored so the promise can be resolved, so that queries can be `await`ed. Would this be useful as documentation of the `enqueueQuery` method>

dslmeinte:
> with some remarks
> 
> as usual, can't judge the TS impl.

I anyway found a bug (and a missing unit test), so there'll be some more work done in this PR.

packages/deltas-websocket/src/client/client-impl.ts
enikao: It is mentioned in the error section there.

dslmeinte:
> > with some remarks
> > as usual, can't judge the TS impl.
> 
> I anyway found a bug (and a missing unit test), so there'll be some more work done in this PR.

Done: turned out to be just a lack of documentation.

PR #229, opened 2025-07-31
Upgrade to littoral-templates@0.5.0

dslmeinte:
+ fix some sh-enanigans and other stuff

### DIFF
--- a/make.sh
+++ b/make.sh
@@ -1,4 +1,5 @@
 #!/bin/sh
+
 # Script to build all packages — it does so *in their dependency order*.
 
 # Have script exit after first failure is detected:
--- a/package-lock.json
+++ b/package-lock.json
@@ -1824,9 +1824,9 @@
       }
     },
     "node_modules/littoral-templates": {
-      "version": "0.4.2",
-      "resolved": "https://registry.npmjs.org/littoral-templates/-/littoral-templates-0.4.2.tgz",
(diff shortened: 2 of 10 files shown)


make.sh
```diff
@@ -1,4 +1,5 @@
 #!/bin/sh
+
```
joswarmer, line 2: It might be easier and shorter to use the npm workspaces by using this:

`npm run build --workspaces --if-present`

joswarmer (approved):
Approve as the only comment has nothing to do with this specific PR.

make.sh
dslmeinte: The thing is: there's a definite order in which the packages should execute their `npm run build` when making from scratch, i.e. with a newly-cloned or `npm run clean`ed repository. The `make.sh` takes care of that order, next to generating some code.

dslmeinte: Depends...I'd say... I had it like that, but then again, I try to do not to too much on each script, so that I can relatively quickly run an `npm run build` to sync up if I made cross-package changes. In the end, `make.sh` does _everything_.

joswarmer: Right! Therefore I usually have a build and a rebuild, where the latter does everything.

dslmeinte: So, you'd have a `build` that just compiles/transpiles/builds, and a `rebuild` that rebuilds everything (essentially from scratch)?

PR #231, opened 2025-08-15
Jos/update validator

joswarmer:
Improved the validator, needed for lionweb-server delta implementation.
When merged, want to publish this as the next beta.


dslmeinte (approved):
I think it'd help overall if we also extracted the definition schema in some way: would be helpful for deltas, commands, events, queries as well.

joswarmer:
> I think it'd help overall if we also extracted the definition schema in some way: would be helpful for deltas, commands, events, queries as well.

Not sure I understand what you want to extract. Do you mean the actual definitions schema's for all delta's? I so, these are currently defined in the lionweb-server project. We might want to rethink where things should go.

PR #258, opened 2025-11-12
Release 0.7.2

dslmeinte:
Note that all commits (except for the last one, setting the version) have been (FF-)merged through a PR, so no further inspection is required. Version 0.7.2 has also already been released.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -226,6 +226,8 @@
         },
         "node_modules/@isaacs/balanced-match": {
             "version": "4.0.1",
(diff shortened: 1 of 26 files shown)


joswarmer:
Since this is already release now, it should be reviewed ok, shouldn't it :-) ?

joswarmer (approved):
Too much has changed since I reviewed the code, makes little sense do to a full review now. And ... it is already released, so it should go to main.,

PR #267, opened 2025-12-03
Fix reference target type to have nullable 'reference' field

### LINKED ISSUE #240: Question about references
Shouldn't reference be nullable here?

https://github.com/LionWeb-io/lionweb-typescript/blob/57c20f9a4552c388628f645c16dfe22603b2ba03/packages/json/src/types.ts#L66-L69

dslmeinte:
Fixes #240


ftomassetti (approved):
I do not see anything immediately wrong

PR #276, opened 2025-12-12
Fix `messageKind` field for request messages of queries to *not* have the `Request` suffix

### LINKED ISSUE #274: Incorrect messageKind for request messages.
As discussed in the meeting from 12-12-2025, we use the technical name of the delta message for the value of “messageKind”, all commands/events do use the technical name as-is, so should t he requests.

So for e.g. the SubscribeToPartitionContents request this should be `messageKind = "SubscribeToPartitionContents"`. 

Thus, the current typescript implementation is incorrect as it does not use the technical name, but `messageKind = "SubscribeToPartitionContentsRequest"` .

dslmeinte:
Fixes #274

### DIFF
--- a/packages/delta-protocol-client/CHANGELOG.md
+++ b/packages/delta-protocol-client/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 0.8.0 — not yet released
 
 * Rework the `LionWebClient` class to use the new `Forest` class from the `class-core` package.
+* Fix `messageKind` field for request messages of queries to *not* have the "`Request`" suffix.
 
 
 ## 0.7.1
--- a/packages/delta-protocol-client/src/client.ts
+++ b/packages/delta-protocol-client/src/client.ts
@@ -205,7 +205,7 @@ export class LionWebClient {
 
     async subscribeToChangingPartitions(queryId: LionWebId, parameters: SubscribeToPartitionChangesParameters): Promise<void> {
         await this.makeQuery({
-            messageKind: "SubscribeToChangingPartitionsRequest",
+            messageKind: "SubscribeToChangingPartitions",
             queryId,
             ...parameters,
             protocolMessages: []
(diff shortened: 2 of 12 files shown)


joswarmer:
> I think it does. So, we just need to add wording to the specification that mentions that the technical name is _not_ necessarily going to be the value of the `messageKind`, owing to sync vs. async?

We are defining properties for messages, so it makes sense to specify the `messageKind` property as well, as we already do for Commands. I think we should do this for Queries as well, certainly if the are implemented using async web-sockets. 

Looking at the three implementations we have (typescript., C#, server), we all have added this exact property. So let's make it official. And explicitly stating that the value of this property is equal to the technical name ensures that various implementations will actually work together, which is the end objective of LionWeb. 

Even then different implementations may use different technical means to send messages (sync and/or async) to each other   , but the message itself is fully specified and identifiable. There is no need to derive the "message kind" from the technical protocol. 

If we do not use the technical name as the value for `messageKind` property, what then is the purpose of this technical name?

Responses have no technical name in the spec, only requests have. We can specify that the technical name (thus the messageKind) of a Response message is the technical name of the Request + "Response"). Or we can specify both technical names explictly.

> I'd argue that's covered by https://lionweb.io/specification/delta/delta-api.html#out-of-scope

In that case, why do we have the `messageKind` property on all Command definitions?

joswarmer:
> > I was not aware we wanted to change that? The [json schema](https://github.com/LionWeb-io/specification/blob/783f6093de63b0e0b438636adb7a990ed2f4b1ed/delta/delta.schema.json#L476) includes the `Request` suffix.
> 
> Isn’t the JSON Schema then not also wrong? Or else I don’t know what was incorrect in the first place.

Yes, it's wrong as well.

dslmeinte:
> Shall we say "The 'spec' for WebSocket binding is whatever is inside the JSON schema"?

Right now, there’s no link directly to the JSON Schema file in the delta section, so the status/role of it is anyway vague. We could make that explicit as: ”we’d do it this way”, and probably mention that `messageKind` equals `<technical name>{Request|Response}` for queries.

dslmeinte:
Dropping this PR: each `Message` type in the TS implementation has the appropriate `messageKind` field, which value is the technical name, with `{Request|Response}` appended for queries.

PR #279, opened 2026-01-16
Extract utilities relying on Node.js to a separate package

dslmeinte:
Also update documentation, and fix a bad link in a badge for the ts-utils package.


pjmolina (approved):
looks good to me

PR #283, opened 2026-01-19
Improve dealing with unresolved references

dslmeinte:
Look at the individual commits for more info

### DIFF
--- a/packages/class-core/src/textualizer.ts
+++ b/packages/class-core/src/textualizer.ts
@@ -40,7 +40,7 @@ type INodeBaseOrNotThere = INodeBase | UnresolvedReference | undefined
 
 const asINodeBases = (value: INodeBaseOrNotThere | INodeBaseOrNotThere[]): INodeBase[] => {
     const isINodeBase = (value: INodeBaseOrNotThere): value is INodeBase =>
-        !(isUnresolvedReference(value) || value === undefined)
+        !(value === undefined || isUnresolvedReference(value))
 
     if (Array.isArray(value)) {
         return value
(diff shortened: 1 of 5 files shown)


packages/class-core-generator/src/entity-types.templates.ts
```diff
@@ -176,9 +169,22 @@ export const typeForLanguageEntity = (imports: Imports) => {
             ]
         }
 
+        const extendsFragment = "extends " + ((superConcept) => {
+            if (isUnresolvedReference(superConcept)) {
+                return `/* unresolved reference to super concept */`
```
enikao: Would it make sense to include the _resolveInfo_ in the message (also at other places)?

packages/build/src/deltas/generator/deserializer-generator.ts
```diff
@@ -30,11 +30,12 @@ import {
     Type
 } from "../definition/Deltas.g.js"
 import { tsTypeForFeatureKind } from "./helpers.js"
+import { isUnresolvedReference } from "@lionweb/core"
 
 const deserializationExpressionForField = (name: string, type: Type) => {
     if (type instanceof FeatureType) {
         const tsMetaType = tsTypeForFeatureKind(type.kind)
-        return `resolved${tsMetaType}From(delta.${name}, ${type.container?.name ?? "<?container?>"}.classifier)`
+        return `resolved${tsMetaType}From(delta.${name}, ${(isUnresolvedReference(type.container) ? undefined : type.container?.name) ?? "<?container?>"}.classifier)`
```
enikao: I would tend to assume in the generators that the languages are valid, and not intermingle checks all over the templates. There are lots of other possible ways to mess up a language, we also don't deal with them in the templates.
We could add a validation run in the default implementation of generating languages.

dslmeinte: This is code that generates code in the `class-core` package: the code generator itself is not user-facing. I baked in the resilience because it helped me develop the generator.

PR #288, opened 2026-02-03
Properly encapsulate LionWeb version 2023.1

dslmeinte:
This is a good thing to do before implementing 2024.1 and 2026.1 — which, oddly enough, is probably much less work and simpler than doing this encapsulation...

The focus really is encapsulating 2023.1, _not parametrizing “everything” with a LionWeb version yet!_

Commits are pretty much atom, but I’ll probably squash-merge this PR anyway.

### DIFF
--- a/packages/core/src/deserializer.ts
+++ b/packages/core/src/deserializer.ts
@@ -5,7 +5,7 @@ import { consoleProblemReporter, ProblemReporter } from "./reporter.js"
 import { MemoisingSymbolTable } from "./m3/symbol-table.js"
 import { Classifier, Containment, Enumeration, Language, PrimitiveType, Property, Reference } from "./m3/types.js"
 import { LionWebVersion } from "./m3/version.js"
-import { defaultLionWebVersion } from "./m3/versions.js"
+import { LionWebVersions } from "./m3/versions.js"
 import { unresolved } from "./references.js"
 import { Node } from "./types.js"
 
@@ -42,7 +42,7 @@ export type DeserializerConfiguration<NT extends Node> = {
     writer: Writer<NT>,
     /**
      * The version of the LionWeb serialization format to deserialize from.
-     * Default = {@link defaultLionWebVersion}.
+     * Default = {@link LionWebVersions.v2023_1}.
      */
     lionWebVersion?: LionWebVersion
     /**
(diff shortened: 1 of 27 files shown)


packages/artifacts/src/generate-artifacts.ts
```diff
@@ -20,13 +20,13 @@ import { languageWithEnum } from "@lionweb/test/dist/languages/with-enum.js"
 import { diagramPath, instancePath, languagePath } from "./paths.js"
 
 
-const lioncore = lioncoreFacade.language
+const lioncore = LionWebVersions.v2023_1.lioncoreFacade.language
```
enikao: Should this be a const? I guess we want to be more flexible about this soon.

packages/build/src/deltas/meta-definition.ts
```diff
@@ -15,11 +15,12 @@
 // SPDX-FileCopyrightText: 2025 TRUMPF Laser SE and other contributors
 // SPDX-License-Identifier: Apache-2.0
 
-import { lioncoreBuiltinsFacade, LanguageFactory } from "@lionweb/core"
+import { LanguageFactory, LionWebVersions } from "@lionweb/core"
 import { concatenator } from "@lionweb/ts-utils"
 
-const { inamed } = lioncoreBuiltinsFacade.classifiers
-const { stringDataType } = lioncoreBuiltinsFacade.primitiveTypes
+const { builtinsFacade } = LionWebVersions.v2023_1
```
enikao: const?

packages/build/src/generate-for-class-core.ts
```diff
@@ -38,7 +38,7 @@ import { getFromHttps } from "./curl.js"
 const inArtifactsPath = (subPath: string) => join("artifacts", subPath)
 
 // generate LionCore-builtins:
-generateLanguage(lioncoreBuiltinsFacade.language, "../class-core/src", { genericImportLocation: "./index.js", header: defaultTrumpfOriginatingApache2_0LicensedHeader })
+generateLanguage(LionWebVersions.v2023_1.builtinsFacade.language, "../class-core/src", { genericImportLocation: "./index.js", header: defaultTrumpfOriginatingApache2_0LicensedHeader })
```
enikao: externalize LionWebVersions?

packages/core/src/deserializer.ts
```diff
@@ -65,15 +72,16 @@ export const deserializerWith = <NT extends Node>(configuration: DeserializerCon
     const symbolTable = new MemoisingSymbolTable(configuration.languages)
     const { writer } = configuration
 
-    const propertyValueDeserializer = configuration.propertyValueDeserializer ?? lioncoreBuiltinsFacade.propertyValueDeserializer
+    const lionWebVersion = configuration?.lionWebVersion ?? defaultLionWebVersion
+    const propertyValueDeserializer = configuration.propertyValueDeserializer ?? lionWebVersion.builtinsFacade.propertyValueDeserializer
```
enikao: Here we're getting the "value converter" from the lionWebVersion. I _think_ we don't do the equivalent in the serializer.

PR #296, opened 2026-02-19
Refactor validation for delta

joswarmer:
Refactor validation to be generic, now all delta messages can be validated as well.
Structure of messages is now in (JSON) objects.


dslmeinte:
> @dslmeinte In the README.md not all packages that are published are mentioned:
> 
> > We (only) release the following packages: core, validation, utilities, cli, class-core, class-core-generator, ts-utils, textgen-utils, io-lionweb-mps-specific
> 
> But `json`, `json-utils` en `json-diff` are also released.

Indeed: could you update the `README.md` for that? (And maybe we can generate that list?)

dslmeinte (changes requested):

dslmeinte (approved):

PR #300, opened 2026-03-13
Update delta definitions from spec

joswarmer:
New definitions files for validating delta messages

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3076,13 +3076,13 @@
             "version": "0.9.0-dev.0",
             "license": "Apache-2.0",
             "dependencies": {
-                "@lionweb/class-core": "0.9.0-beta.0",
-                "@lionweb/class-core-generator": "0.9.0-beta.0",
-                "@lionweb/core": "0.9.0-beta.0",
-                "@lionweb/json": "0.9.0-beta.0",
-                "@lionweb/textgen-utils": "0.9.0-beta.0",
-                "@lionweb/ts-utils": "0.9.0-beta.0",
-                "@lionweb/utilities": "0.9.0-beta.0",
+                "@lionweb/class-core": "0.9.0-beta.1",
+                "@lionweb/class-core-generator": "0.9.0-beta.1",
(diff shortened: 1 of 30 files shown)


packages/validation/src/validators/definitions/CommandDefinitions.ts
```diff
@@ -1273,12 +1308,26 @@ export const CommandDefinitions: MessageGroup = {
             ],
         },
         {
-            name: "CompositeCommand",
```
dslmeinte, line 1276: As far as I’m aware, we reinstated `CompositeCommand` (and `CompositeEvent`), while this is a replacement with `ChunkedCommand` (and `ChunkedEvent`).

Also, there’s no support for `Chunked{Command|Event}` in the other packages yet. I don’t mind beta-releasing just this PR, though.

package-lock.json
dslmeinte, line 1: This commit is seemingly not necessary: even deleting this file, and running `npm i` doesn’t produce this diff on my machine.

packages/validation/src/validators/definitions/CommandDefinitions.ts
joswarmer: CompositeCommand is still there, just at the very end of this file. Git just gets confused and show it as added there.  So this adds ChunkedCommand.

This won't be a problem because the chunkedFlag is optional, so any command not implementing it will still be considered correct.

package-lock.json
joswarmer: I must be mistaken. Thought I had the local npm registry in this file again, but presumably not checked in.

packages/validation/src/validators/definitions/CommandDefinitions.ts
joswarmer: So ... are you ok with merging this into development?

dslmeinte: (Made an issue for support for chunked stuff elsewhere in this codebase: https://github.com/LionWeb-io/lionweb-typescript/issues/301)

PR #306, opened 2026-03-24
added delta test for references to M2 elements

enikao:


dslmeinte (approved):
But please update CHANGELOG!

PR #307, opened 2026-03-25
Add extra function

joswarmer:
Just adding an extra function.

### DIFF
--- a/packages/json-utils/src/LionWebJsonChunkWrapper.ts
+++ b/packages/json-utils/src/LionWebJsonChunkWrapper.ts
@@ -25,6 +25,17 @@ export class LionWebJsonChunkWrapper {
     protected nodesIdMap: Map<LionWebId, LionWebJsonNode> = new Map<LionWebId, LionWebJsonNode>()
 
     /**
+     * Create from a collection of nodes
+     * @param nodes
+     */
+    static fromNodesArray(nodes: LionWebJsonNode[]) {
+        return new LionWebJsonChunkWrapper({
+            languages: [],
+            serializationFormatVersion: "2023.1",
+            nodes: nodes
+        })
+    }
+    /**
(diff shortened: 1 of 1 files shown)


packages/json-utils/src/LionWebJsonChunkWrapper.ts
```diff
@@ -69,8 +80,68 @@ export class LionWebJsonChunkWrapper {
...
+        }
+    }
+
     /**
-     * Return the target nodes inside `reference` as a list of actual nodes (LionWebJsonNode[])
+     * Return an array with _all_ child nodes of `node`
+     * @param nodeId
+     */
+    getAllChildNodes(node: LionWebJsonNode): LionWebJsonNode[] {
+        return node.containments?.flatMap(cont => this.getContainmentNodes(node, cont.containment))
```
dslmeinte: The `?` should not be necessary, right? (As it wasn’t above, e.g.)

packages/json-utils/src/LionWebJsonChunkWrapper.ts
```diff
@@ -69,8 +80,68 @@ export class LionWebJsonChunkWrapper {
...
+    }
+
+    /**
+     * Return an array with all child nodes of `node`
```
dslmeinte: ...for a particular containment.

dslmeinte (approved):
But see remarks!

packages/json-utils/src/LionWebJsonChunkWrapper.ts
joswarmer: Indeed, and in line 132 as well, fixed both.

packages/json-utils/src/LionWebJsonChunkWrapper.ts
joswarmer: Fixed

Issue #308, opened 2026-04-03
Avoid name collisions in code generated by the class-core generator


dslmeinte:
Name collisions occur when:

- Concepts have names that coincide with names of types coming from `@lionweb/[class-]core` or `@lionweb/json`, e.g. `Class`, `Node`, `Classifier`, etc.. We should be able to solve this using *-namespaces with names that are illegal as LionWeb concept names.
- Features have names that coincide with properties of `NodeBase`. It’s probably best (but a breaking change!) to put LionWeb-internal properties in a separate sub object that has a name that’s illegal as a LionWeb feature name.

Alternatively, we can *escape*/“sanitize” colliding concept and feature names. (I don’t know yet what’d be the best solution.)

joswarmer:
Not sure what you mean, I cannot find `Node` or `Classifier` or `Class` in `@lionweb/json`.

dslmeinte:
> Not sure what you mean, I cannot find `Node` or `Classifier` or `Class` in `@lionweb/json`.

All these names occur as types in `@lionweb/core`. But, basically, if you were to generate code for LionCore (+builtins) using the class-core generator, the result should compile (without having to modify it), right?

joswarmer:
If you are afraid of duplicate class/type/interface names, that isn't a problem.  The user of these classes can always use `import as` to distinguish them.

File Node.ts:
```typescript
export class Node {
    node: string
}
``` 
File MyNode.ts
```typescript
export class Node {
    mynode: string
}
``` 
Usage
```typescript
import { Node as HisNode } from "./Node.js"
import { Node as MyNode } from "./MyNode.js"

const his = new HisNode()
const my = new MyNode()

his.node
my.mynode
```

dslmeinte:
The point is that the class-core generator has to use import aliasing to avoid collisions. (Name sanitization has the drawback that names in generated code differ from the ones in the LionWeb language, but is anyway necessary to deal with names that were never going to be valid identifiers in any reasonable GPL.)

dslmeinte:
OK, name sanitization is not necessary after all; see here: https://github.com/LionWeb-io/specification/issues/48

And that means that we can use a `$` prefix to do import aliasing and avoid collisions.

joswarmer:
Using import * as ... allows you to tuse the original name (with a prefix), like:
```typescript
import * as LionWeb from "./Node.js"
```
Usage
```typescript
const node = new LionWeb.Node()
```

[...]

PR #309, opened 2026-04-06
Avoid collisions between types in generated code and types imported from LionWeb itself

dslmeinte:
Fixes issue #308


pjmolina (approved):
looks good to me

PR #316, opened 2026-04-20
Jos/fix diff update delta definitions

joswarmer:
Update all delta definitions.
Fix a small error in the diff.

### DIFF
--- a/packages/json-diff/CHANGELOG.md
+++ b/packages/json-diff/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.10.0
+
+Fixing bug in diffing references.
+
 ## 0.9.0, 0.8.0, 0.7.2
 
 (No changes)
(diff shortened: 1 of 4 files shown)


dslmeinte (approved):
Looks good!

Two remarks:

- Please update `CHANGELOG.md`s.
- We (/I) have to still propagate this to other delta-related packages.

joswarmer:
> Looks good!
> 
> Two remarks:
> 
> * Please update `CHANGELOG.md`s.
> * We (/I) have to still propagate this to other delta-related packages.

I will update the changelog.  We do need to update other packages, but I do need a new beta release somewhere this week for use with the LionWeb-server and Freon. 
Note that the changes are only the definitions used by the validator, not the typescript types/interfaces.
There are separate typescript types defined in the lionweb-server (we need to unify these at some point in time).

PR #317, opened 2026-04-20
Add old* properties to various commands

dslmeinte:
As a result of: https://github.com/LionWeb-io/specification/pull/468

### DIFF
--- a/packages/delta-protocol-common/src/payload/command-types.ts
+++ b/packages/delta-protocol-common/src/payload/command-types.ts
@@ -161,6 +161,7 @@ export interface MoveAndReplaceChildInSameContainmentCommand extends Command {
     newIndex: number
     oldIndex: number
     replacedChild: LionWebId
+    movedChild: LionWebId
 }
 
 /** § 5.7.6.1 */
--- a/packages/delta-protocol-common/src/payload/command-types.ts
+++ b/packages/delta-protocol-common/src/payload/command-types.ts
@@ -101,24 +101,30 @@ export interface MoveChildFromOtherContainmentCommand extends Command {
     newParent: LionWebId
     newContainment: LionWebJsonMetaPointer
     newIndex: number
+    oldParent: LionWebId
+    oldContainment: LionWebJsonMetaPointer
+    oldIndex: number
     movedChild: LionWebId
 }
 
 /** § 5.7.5.5 */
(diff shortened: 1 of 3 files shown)


packages/delta-protocol-common/src/payload/command-types.ts
```diff
@@ -128,23 +134,32 @@ export interface MoveAndReplaceChildFromOtherContainmentCommand extends Command
...
     newContainment: LionWebJsonMetaPointer
     newIndex: number
+    oldContainment: LionWebJsonMetaPointer
+    oldIndex: number
     replacedChild: LionWebId
     movedChild: LionWebId
 }
 
 /** § 5.7.5.9 */
 export interface MoveAndReplaceChildInSameContainmentCommand extends Command {
```
joswarmer, line 157: Missing movedChild property

packages/delta-protocol-common/src/translators/delta-to-command.ts
```diff
@@ -195,21 +201,30 @@ export const deltaToCommandTranslator = (
...
                 replacedChild: delta.replacedChild.id,
                 movedChild: delta.movedChild.id
             })
         }
         if (delta instanceof ChildMovedAndReplacedInSameContainmentDelta) {
             return completed<MoveAndReplaceChildInSameContainmentCommand>("MoveAndReplaceChildInSameContainment", { // § 5.7.5.9
+                parent: delta.parent.id,
+                containment: metaPointerFor(delta.containment),
                 newIndex: delta.newIndex,
+                oldIndex: delta.oldIndex,
```
joswarmer, line 227: Missing movedChild

joswarmer: Since movedChild is a property of `ChildMovedAndReplacedInSameContainmentDelta`, shouldn't typescript give a type error here?

joswarmer (approved):
Missing movedChild in one Command and one Delta.

erk-dik (approved):

packages/delta-protocol-common/src/translators/delta-to-command.ts
dslmeinte: Not until you put the `movedChild` property in the `MoveAndReplaceChildInSameContainmentCommand` type. The TS compiler can’t know that all properties of `ChildMovedAndReplacedInSameContainmentDelta` should be accessed.

PR #318, opened 2026-04-22
Allow idempotent setting of a single containment’s value

dslmeinte:


packages/class-core-test/src/value-managers/containment_mixed.tests.ts
```diff
@@ -70,8 +71,8 @@ describe("containment (mixed)", () => {
         const newLTC = (id: LionWebId)=>
             LinkTestConcept.create(id, receiveDelta);
         const topLTC = newLTC("a");
-        const partition = TestPartition.create("partition", receiveDelta)
-        partition.addLinks(topLTC)
+        const partition = TestPartition.create("partition", receiveDelta);
```
pjmolina, line 74: Just curiosity: Added semicolons finally?
Some lines have it, someothers not.
A linter can enforce the consistency on this.

pjmolina (approved):
Approved. Small comment about semicolons. `;`

packages/class-core-test/src/value-managers/containment_mixed.tests.ts
dslmeinte: I enforced the consistency manually here. But I would rather remove _all_ semicolons, but I’d do that in a big chore PR.

PR #319, opened 2026-04-22
Implement splittable messages

dslmeinte:
+ handle ErrorResponse for queries
  + extract query message kinds
 
 Implementation is quite clean, but not entirely trivial.

### DIFF
--- a/packages/delta-protocol-client/src/chunking.ts
+++ b/packages/delta-protocol-client/src/chunking.ts
@@ -23,7 +23,7 @@ import {
     Message,
     SplittableMessage
 } from "@lionweb/delta-protocol-common"
-import { LionWebJsonChunk } from "@lionweb/json"
+import { LionWebJsonChunk, LionWebJsonDeltaChunk } from "@lionweb/json"
 
 
 /**
@@ -36,7 +36,7 @@ export class ChunkedInfo {
     /**
      * The received chunks, indexed by their `continuedChunkSequenceNumber`.
      */
-    private readonly chunks: LionWebJsonChunk[] = []
+    private readonly chunks: LionWebJsonDeltaChunk[] = []
     /**
      * The number of unique received chunks.
      * Note can be different from `chunks.length` because of `undefined` entries and (theoretically) doubly-received chunks.
@@ -93,7 +93,6 @@ export class ChunkedInfo {
         this.chunks.forEach((continuedChunk) => {
             mergedChunk.nodes.push(...continuedChunk.nodes)
         })
-        // TODO  also do something with languages?
         return this.initialMessage
     }
 
--- a/packages/delta-protocol-client/src/chunking.ts
+++ b/packages/delta-protocol-client/src/chunking.ts
@@ -1,0 +1,141 @@
+// Copyright 2026 TRUMPF Laser SE and other contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License")
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-FileCopyrightText: 2026 TRUMPF Laser SE and other contributors
+// SPDX-License-Identifier: Apache-2.0
+
+import {
+    ContinuedChunkMessage,
+    Event,
+    isChunkedEvent,
+    maybeChunkPropertyForSplittableEvent,
+    Message,
+    SplittableMessage
+} from "@lionweb/delta-protocol-common"
+import { LionWebJsonChunk } from "@lionweb/json"
+
+
+/**
+ * Instances of this class keep track of a *split message*.
+ * (See § 3.7.1 of the specification of the delta protocol.)
+ * Its state is updated through its {@link maybeCompletedMessage} method.
+ */
+export class ChunkedInfo {
+
+    /**
+     * The received chunks, indexed by their `continuedChunkSequenceNumber`.
+     */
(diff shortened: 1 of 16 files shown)


dslmeinte:
I might add some T.-specific tests still.

pjmolina (approved):
Looks good to me

packages/delta-protocol-client/src/chunking.ts
```diff
@@ -0,0 +1,141 @@
...
+ * Instances of this class keep track of a *split message*.
+ * (See § 3.7.1 of the specification of the delta protocol.)
+ * Its state is updated through its {@link maybeCompletedMessage} method.
+ */
+export class ChunkedInfo {
+
+    /**
+     * The received chunks, indexed by their `continuedChunkSequenceNumber`.
+     */
+    private readonly chunks: LionWebJsonChunk[] = []
```
joswarmer: According to the specification the chunks in chunked messages are DeltaChunks, without "languages", "serializationVersion" etc.
Should have a separate type for this (the lionweb-server certainly has).

packages/delta-protocol-client/src/chunking.ts
```diff
@@ -0,0 +1,141 @@
...
+        if (this.numberOfReceivedChunks !== this.lastChunkSequenceNumber + 1) {
+            return undefined
+        }
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        const mergedChunk = (this.initialMessage as any)[this.chunkProperty] as LionWebJsonChunk
+        this.chunks.forEach((continuedChunk) => {
+            mergedChunk.nodes.push(...continuedChunk.nodes)
+        })
+        // TODO  also do something with languages?
+        return this.initialMessage
```
joswarmer, line 96: Not if there is a correct type for chunk (see earlier comment).

packages/delta-protocol-client/src/chunking.ts
```diff
@@ -0,0 +1,141 @@
...
+    maybeCompletedMessage(message: ContinuedChunkMessage): (Message | undefined) {
+        const { chunk, continuedChunkCompleted, continuedChunkSequenceNumber } = message
+        if (this.numberOfReceivedChunks === this.lastChunkSequenceNumber + 1) {
+            throw new Error(`split message already completed`)
+        }
+        if (continuedChunkCompleted) {
+            if (this.lastChunkSequenceNumber === -Infinity) {
+                this.lastChunkSequenceNumber = continuedChunkSequenceNumber
+            } else {
+                throw new Error(`continued chunk sequence declared complete more than once`)
```
joswarmer, line 79: This should be added as an error in the specifciation, as the other errors in here as well.

packages/delta-protocol-client/src/chunking.ts
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    /**
+     * Updates this {@link ChunkInfo} with the given {@link ContinuedChunkMessage}, and
+     * @returns the completed message if the chunking is complete,
+     * or `undefined` if it isn’t.
+     */
+    maybeCompletedMessage(message: ContinuedChunkMessage): (Message | undefined) {
+        const { chunk, continuedChunkCompleted, continuedChunkSequenceNumber } = message
+        if (this.numberOfReceivedChunks === this.lastChunkSequenceNumber + 1) {
```
joswarmer, line 72: Wouldn't it be more clear to keep the `continuedChunkCompleted` flag in the ChunkInfo?  It took me quite some puzzling to figure out that the comparison in this if meant that the chunk was already complete.

The error check below would also become much easier.

packages/delta-protocol-client/src/chunking.ts
dslmeinte: Hmmm, I can’t find that type in the specification! It would make sense, though.

packages/delta-protocol-client/src/chunking.ts
dslmeinte: The various chunks of a chunked message could arrive out of order, so `continuedChunkCompleted` merely means that we now know the total number of chunks for this split message, but other chunks might still arrive. (At least, that’s how I read the specification.)

joswarmer: Ah! I assume that the messages always come in the correct sequence (as sent by the client). I think this generally must be true for the communication channel, otherwise delta's might not make sense. Not sure we specify this explicitly, but we should.
The fact that the continuedChunks messages should be sent in the correct order is also an assumption, and I think we should add this to the spec.

dslmeinte: I’m not sure how explicit we mention it, but my working assumption has been all along that events can arrive out-of-order. (That’s why I’m using a queue to store out-of-order events.) So, let’s check whether the spec. is explicit (enough) about it.

packages/delta-protocol-client/src/chunking.ts
dslmeinte: (Discussed on Fri April 24th.)

packages/delta-protocol-client/src/chunking.ts
dslmeinte: (Discussed on Fri April 24th.)

packages/delta-protocol-client/src/chunking.ts
dslmeinte: Will do in a PR on the `specification` repo.

packages/delta-protocol-client/src/chunking.ts
dslmeinte: Will add text in a PR on `specification`.

joswarmer: Interesting! If events arrive out of order, even one single client will not work correctly using the delta protocol with the server. E.g. the client send:
- `AddChild`: Add Child (anywhere) with node id = A
- `AddProperty`: Add property to node with id A

If these arrive not in order, and the `AddProperty` comes first, this commands will be incorrect (node with ID A does not exist).

My conclusion is that if the order received is different from the order sent the whole delta protocol will, never work. Waiting for the PR on specification to review :-).

PR #324, opened 2026-04-28
Provide (some) support for custom messages

### LINKED ISSUE #311: Implement custom messages of delta protocol
It seems that there’s not much to concretely support in `@lionweb/delta-protocol-*` other than some generic types.

dslmeinte:
Fixes issue #311

### DIFF
--- a/packages/delta-protocol-common/src/payload/common.ts
+++ b/packages/delta-protocol-common/src/payload/common.ts
@@ -16,6 +16,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import { LionWebId, LionWebJsonChunk } from "@lionweb/json"
+import { isValidIdentifier } from "@lionweb/core"
 
 export interface Message {
     messageKind: string
@@ -53,4 +54,22 @@ export interface ContinuedChunkMessage extends Message {
     continuedChunkCompleted: boolean
     continuedChunkSequenceNumber: number
 }
+
+
+/**
+ * Type def. for custom message kinds, meaning:
+ *  - MUST adhere to the same format as identifiers,
+ *  - MUST start with "custom_", and
+ *  - MUST have at least 8 characters.
+ * (See §5.3.)
+ *
+ * *Note*: use the {@link isValidCustomMessageKind} to actually check correctness.
+ */
+export type CustomMessageKind = `custom_${string}`
+
+/**
+ * @return Whether the given `messageKind` string is a valid message kind string.
+ */
+export const isValidCustomMessageKind = (messageKind: string): messageKind is CustomMessageKind =>
+    isValidIdentifier(messageKind) && messageKind.startsWith("custom_") && messageKind.length >= 8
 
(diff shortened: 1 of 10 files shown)


packages/delta-protocol-common/src/payload/query-types.ts
```diff
@@ -27,6 +27,37 @@ export interface QueryMessage extends DeltaAdditionalInfo {
...
+export interface CustomQueryResponse extends QueryMessage {
+    /*
+     * messageKind to be specified:
+     *  - MUST adhere to the same format as identifiers,
+     *  - MUST start with "custom_", and
+     *  - MUST have at least 8 characters,
+     *  - SHOULD end in "Response".
+     */
+}
+
```
joswarmer, line 53: Since all of these interfaces are _very much_ empty, they are completely useless within TypeScript. 
What is the purpose of adding them?

dslmeinte: Just a hint for implementors that they can extend these interfaces, and then should adhere to the specifics of the spec.

packages/delta-protocol-common/src/payload/event-types.ts
```diff
@@ -37,6 +37,21 @@ export interface OriginatedEvent extends Event {
 }
 
 
+/**
+ * “Abstract” interface for custom commands.
```
joswarmer: Typo: commands => events

packages/delta-protocol-common/src/payload/query-types.ts
joswarmer: Even so, since the interfaces have no properties, the type-checking from TS will find `CustomCommand` and `Command` identical, so extending these interfaces does not provide implementors with any help.

joswarmer (approved):

packages/delta-protocol-common/src/payload/query-types.ts
dslmeinte: With the 2nd commit, I added some help.

packages/delta-protocol-common/src/payload/common.ts
```diff
@@ -54,3 +55,21 @@ export interface ContinuedChunkMessage extends Message {
...
+/**
+ * Type def. for custom message kinds, meaning:
+ *  - MUST adhere to the same format as identifiers,
+ *  - MUST start with "custom_", and
+ *  - MUST have at least 8 characters.
+ * (See §5.3.)
+ *
+ * *Note*: use the {@link isValidCustomMessageKind} to actually check correctness.
+ */
+export type CustomMessageKind = `custom_${string}`
```
joswarmer, line 68: Interesting what you can do with types in typescript!

joswarmer: Just a remark (on the spec actually): all message kinds start with an uppercase character, maybe custom messages should do this as well.

dslmeinte: I’ll made an issue in the `specification` repo for that: https://github.com/LionWeb-io/specification/issues/476

PR #325, opened 2026-05-05
Feature/dictionary validation

joswarmer:
Add dictionary validation and type of AdditionalInfo is Dictionary.


dslmeinte (changes requested):
`typeof object === "object" && !Array.isArray(object)` alone doesn’t guarantee that all keys and values are strings, and that all keys are valid LionWeb IDs.

packages/validation/src/validators/definitions/DeltaTypesDefinitions.ts
```diff
@@ -56,8 +60,8 @@ export const DeltaTypesDefinitions: TypeGroup = {
                 },
                 {
                     name: "data",
-                    type: "KeyValuePair",
```
dslmeinte, line 59: The structured type `KeyValuePair` can then also be deleted, assuming that `ResponseMessage` (which `data` property uses `KeyValuePair`) is unused.

packages/validation/src/validators/generic/SyntaxValidator.ts
```diff
@@ -170,6 +170,11 @@ export class SyntaxValidator {
             this.validationResult.issue(new Syntax_PropertyTypeIssue(jsonContext, propertyName, propDef.primitiveType, typeof object))
             return false
         }
+        // typof array is object in JS, so check for this special (weird) case
```
dslmeinte: typo: "typeof"

packages/validation/src/validators/generic/SyntaxValidator.ts
```diff
@@ -182,6 +180,40 @@ export class SyntaxValidator {
...
+     * @param jsonContext   The context in the JSON
+     */
+    validateDictionary(propertyName: string, propDef: PrimitiveType, object: unknown, jsonContext: JsonContext): boolean {
+        if (typeof object === "object" && !Array.isArray(object) && object !== null) {
+            for (const key of Object.keys(object)) {
+                if (typeof key !== "string") {
+                    this.validationResult.issue(
+                        new GenericIssue(jsonContext, `property '${propertyName}' key '${key}' should be a string, it is a '${typeof key}'`)
+                    )
+                    return false
```
dslmeinte, line 197: Please also check whether the key is an ID-compatible string (as per spec.)

packages/validation/src/validators/generic/SyntaxValidator.ts
```diff
@@ -182,6 +180,40 @@ export class SyntaxValidator {
...
+                }
+                const value = (object as any)[key]
```
dslmeinte: Prepend with the following line to fix the linting issue:
```
// eslint-disable-next-line @typescript-eslint/no-explicit-any
```

dslmeinte (changes requested):

dslmeinte:
Also: please update `CHANGELOG.md`

packages/validation/src/validators/definitions/DeltaTypesDefinitions.ts
joswarmer: I have left this in, because it is still used in the bulk api and the lionweb-server uses this validator to validate bulk messages.
This api is not updated yet to all the changes that result from specifying the delta api. (TODO!)
Will cleanup this when updating the bulk API

packages/validation/src/validators/generic/SyntaxValidator.ts
joswarmer: fixed

joswarmer:
> `typeof object === "object" && !Array.isArray(object)` alone doesn’t guarantee that all keys and values are strings, and that all keys are valid LionWeb IDs.

Added checks for this in the validator

dslmeinte (approved):

packages/validation/src/validators/generic/SyntaxValidator.ts
joswarmer: Indeed, added.

PR #338, opened 2026-06-22
Update to offset based moves in delta protocol impl

dslmeinte:


packages/validation/src/validators/definitions/CommandDefinitions.ts
```diff
@@ -618,7 +618,7 @@ export const CommandDefinitions: MessageGroup = {
                     mayBeNull: false,
                 },
                 {
-                    name: "newIndex",
+                    name: "indexOffset",
```
joswarmer, line 621: I always generate this file from the Delta LionWeb model, so it is automatically in sync with other generated files in the server project. 
But ... I did not realize these property names were changed, I will update the Delta model accordingly later this week. Should result in the same TS code.

packages/validation/src/validators/definitions/CommandDefinitions.ts
```diff
@@ -1120,7 +1120,7 @@ export const CommandDefinitions: MessageGroup = {
...
```
joswarmer, line 1123: See previous comment.

packages/validation/src/validators/definitions/EventDefinitions.ts
```diff
@@ -1337,7 +1337,7 @@ export const EventDefinitions: MessageGroup = {
                     mayBeNull: false,
                 },
                 {
-                    name: "newIndex",
+                    name: "indexOffset",
```
joswarmer, line 1340: Idem

packages/validation/src/validators/definitions/EventDefinitions.ts
```diff
@@ -737,7 +737,7 @@ export const EventDefinitions: MessageGroup = {
...
```
joswarmer, line 740: Idem

joswarmer (approved):

packages/validation/src/validators/definitions/CommandDefinitions.ts
dslmeinte: Where does that Delta LionWeb model “live” again?

PR #341, opened 2026-06-29
Do not fire on unregistered partitions (+ a bug fix)

dslmeinte:


packages/class-core-test/src/forest.tests.ts
```diff
@@ -176,6 +182,33 @@ describe("Forest", () => {
         deepEqual(deltas, [new PartitionAddedDelta(partition)]) // no partition deleted delta re-emitted
     })
 
+    it("adding a populated partition forest only fires 1 delta", () => {
```
enikao, line 185: Maybe add another test: create a deep node and add it as a child to a non-partition; should still emit only one delta

enikao (approved):

PR #344, opened 2026-07-10
Fixes w.r.t. the delta protocol

dslmeinte:


packages/class-core/src/value-managers/containments.ts
```diff
@@ -394,7 +394,7 @@ export abstract class MultiContainmentValueManager<T extends INodeBase> extends
         const participants = this.moveAndReplaceOffsetBasedDirectly(oldIndex, indexOffset);
         if (participants !== undefined) {
             const [movedChild, replacedChild] = participants;
-            this.emitDelta(() => new ChildMovedAndReplacedInSameContainmentDelta(this.container, this.containment, oldIndex, indexOffset, replacedChild, movedChild));
+            this.emitDelta(() => new ChildMovedAndReplacedInSameContainmentDelta(this.container, this.containment, oldIndex, indexOffset, movedChild, replacedChild));
```
enikao, line 397: Does this change the signature or fixes an implementation bug? The former would feel dangerous.

dslmeinte: This fixes an implementation bug: the call now aligns with that class’s constructor signature: `constructor(parent: INodeBase, containment: Containment, oldIndex: number, newIndex: number, movedChild: INodeBase, replacedChild: INodeBase);`

enikao (approved):

PR #354, opened 2026-07-27
Fix&rework class-core textualizer for widened SingleRef type

dslmeinte:
+ Emit `UnresolvedReference` instances for unresolvable references.


pjmolina (approved):
looks good to me

Issue #359, opened 2026-08-17
Chunks in events and commands should be DeltaChunks


joswarmer:
The various chunks in events and commands in `delta-protocol-common/src/payload ` have `LionWebJsonChunk` as type, but it should be the LionWeb Delta Chunk type, without the top `languages` and `serializationFormatVersion` properties.

dslmeinte:
Fixed through PR #360

PR #360, opened 2026-08-17
Fix issue #359

dslmeinte:


pjmolina (commented):
looks good to me

dslmeinte:
> looks good to me

I made one small change, and I’d like Jos to have a look at this before I merge, which he can have tomorrow earliest.

packages/json/src/types.ts
```diff
@@ -36,10 +36,16 @@ export type LionWebJsonMetaPointer = {
     key: LionWebKey
 }
 
-export type LionWebJsonChunk = {
+/**
+ * A type to DRY {@link LionWebJsonChunk} and {@link LionWebJsonDeltaChunk}.
```
joswarmer, line 40: This is of course technically correct, but this looks very over-engineered to me.  And it makes the code just more verbose. more complex and  harder to read.

joswarmer (approved):
Looks good.

packages/json/src/types.ts
dslmeinte: But I do need the type in the sense that the class-core deserializer needs to be able to just work on nodes, without the `languages` and `serializationFormatVersion` fields (in the same object). The alternative would be to rely on structural typing to achieve the same, but that would be similarly verbose.

joswarmer: The type `LionWebJsonDeltaChunk` is there without needing the `OnlyNodesOfLionWebJsonChunk`. Why do you then need the `OnlyNodesOfLionWebJsonChunk` type as well?

dslmeinte: Euh, `LionWebJsonDeltaChunk` is a direct copy of `OnlyNodesOfLionWebJsonChunk`!

joswarmer: I don't think you understand what I am trying to say.  I would just write:

```
export type LionWebJsonDeltaChunk = {
    nodes: LionWebJsonNode[]
}

export type LionWebJsonChunk {
    serializationFormatVersion: LionWebSerializationFormatVersion
    languages: LionWebJsonUsedLanguage[]
    nodes: LionWebJsonNode[]
}
```
No need for the `OnlyNodesOfLionWebJsonChunk & `.

dslmeinte: But I’d still need the `OnlyNodesOfLionWebJsonChunk` type (or its structural equivalent) in the `Deserializer` type from the `class-core` package.

joswarmer: You can use `LionWebJsonDeltaChunk` for this.

dslmeinte: Yeah, but that doesn’t accurately reflect intent.

dslmeinte: So, I’d propose to leave things as they are now, and I’ll just go ahead and merge.

PR #365, opened 2026-08-27
Deserialization of deltas

dslmeinte:
This fixes a problem surfaced by the `lionweb-integration-testing` repo in this commit: https://github.com/LionWeb-io/lionweb-typescript/pull/365/changes/5a19ac4ae3b5abd7706f55442504404fa9442955

(commits are atomic)

### DIFF
--- a/packages/delta-protocol-common/src/translators/event-to-delta.ts
+++ b/packages/delta-protocol-common/src/translators/event-to-delta.ts
@@ -34,6 +34,7 @@ import {
     ChildReplacedDelta,
     CompositeDelta,
     Deserializer,
+    DetailedDeserialization,
     IDelta,
     IdMapping,
     IdOrNull,
@@ -47,8 +48,7 @@ import {
     PropertyDeletedDelta,
     ReferenceAddedDelta,
     ReferenceChangedDelta,
-    ReferenceDeletedDelta,
-    RootsWithIdMapping
+    ReferenceDeletedDelta
 } from "@lionweb/class-core"
 import { featureResolversFor, LionWebVersions, PropertyValueDeserializer, referenceToSet } from "@lionweb/core"
 import { LionWebJsonDeltaChunk } from "@lionweb/json"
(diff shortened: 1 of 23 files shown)


packages/delta-protocol-common/src/translators/event-to-delta.ts
```diff
@@ -151,24 +156,26 @@ export const eventToDeltaTranslator = (
...
             case "ChildAdded": { // § 5.8.5.1
-                const { parent, newChild, containment, index } = event as ChildAddedEvent
-                const resolvedNode = idMapping.nodeBaseFromId(parent)
-                const resolvedContainment = resolvedContainmentFrom(containment, resolvedNode.classifier)
-                return new ChildAddedDelta(resolvedNode, resolvedContainment, index, deserializedNodeFrom(newChild))
+                const { parent, containment, index } = event as ChildAddedEvent
+                const resolvedParent = idMapping.nodeBaseFromId(parent)
+                const resolvedContainment = resolvedContainmentFrom(containment, resolvedParent.classifier)
+                return new ChildAddedDelta(resolvedParent, resolvedContainment, index, deserializeAsSingleDeltaChunk<ChildAddedEvent>("newChild"))
+                // (It doesn’t matter here whether the anchor node in newChild actually declares parent as its `parent`.)
```
enikao: That's not what the spec says: https://lionweb.io/specification/delta/delta-api.html#cmd-AddChild

> newChild 	
> 
> [single chunk](https://lionweb.io/specification/delta/delta-api.html#singleChunkType) Subtree to form a new child. The anchor node MUST contain the [parent](https://lionweb.io/specification/delta/delta-api.html#AddChild.parent) as parent.[[56](https://lionweb.io/specification/delta/delta-api.html#_footnotedef_56)] All nodes in that subtree MUST be [new nodes](https://lionweb.io/specification/delta/delta-api.html#new-nodes).

dslmeinte: The serialized anchor node must declare parent as its parent, but the implementation doesn’t check for that. The implementation also doesn’t check for new-ness — meaning that none of the IDs are in `idMapping`. Do you think it’s a good idea to check there?

btw: Support for shallow chunks in `PartitionAdded` is also slightly hairy, because it essentially borders on bulk protocol semantics.

enikao: I fiddled with both points in C#.

Regarding checking, I don't have a rigid approach. Some things are checked, others not, others as side-effect.
I guess once we get more specific about validity we can flesh this out.

I tried several variants how to handle shallow partitions, none of them felt really clean. I wouldn't go for bulk merge semantics, that's very hard to understand.

enikao: Maybe adjust the comment to clarify the distinction between what the spec says and what practically matters atm.

enikao (approved):

PR #367, opened 2026-09-02
Fix bug in insertAtIndex method of multi-valued containment managers

dslmeinte:
Quasi-courtesy PR: will merge this evening at the latest.

### DIFF
--- a/packages/class-core-test/src/value-managers/containment_0_1.tests.ts
+++ b/packages/class-core-test/src/value-managers/containment_0_1.tests.ts
@@ -152,7 +152,7 @@ describe("write access to a [0..1] containment through .set", () => {
         deepEqual(deltas[2], new ChildMovedFromOtherContainmentInSameParentDelta(parent, testLanguageBase.LinkTestConcept_containment_1, 0, child, testLanguageBase.LinkTestConcept_containment_0_1, 0));
     });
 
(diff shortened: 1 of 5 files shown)


packages/class-core-test/src/value-managers/containment_0_1.tests.ts
```diff
@@ -152,7 +152,7 @@ describe("write access to a [0..1] containment through .set", () => {
         deepEqual(deltas[2], new ChildMovedFromOtherContainmentInSameParentDelta(parent, testLanguageBase.LinkTestConcept_containment_1, 0, child, testLanguageBase.LinkTestConcept_containment_0_1, 0));
     });
 
-    it("moving a child within same containment (no ChildMovedInSameContainmentDelta)", () => {
+    it("assigning a child idempotently", () => {
```
enikao, line 155: Is the change here and in `containment_1.tests.ts` on purpose?
_Might_ be a mixup, can't judge.

enikao (approved):

packages/class-core-test/src/value-managers/containment_0_1.tests.ts
dslmeinte: Is on purpose: just improving test titles — (see also 2nd line in the message for that commit)

Issue #369, opened 2026-09-08
Have the `class-core-generator` generate type guard functions for non-concrete classifiers


dslmeinte:
This is convenient especially for interfaces, because you can’t do `<instance> instanceof <interface type>`.

For abstract classes (that correspond to abstract concepts) the problem doesn’t really exist, but for consistency/symmetry we could also generate those, plus then we can just rely on recursion.

Where would be put these type guard functions?

PR #370, opened 2026-09-08
Type guard functions in `class-core`-generated language bases

dslmeinte:
Solves issue #369.

(Commits are “atomic”.)


packages/class-core-generator/src/helpers/classifiers.ts
```diff
@@ -63,3 +65,33 @@ export const featuresToConcretelyImplementOf = (classifier: Classifier): Feature
         .filter((feature) => implementedFeatures.indexOf(feature) === -1)
 }
 
+
+/**
+ * A type alias for a {@link Map} mapping {@link Classifier classifiers} to their sub classifiers.
```
enikao: Technically, these are "specializations", correct?

I tend to actually use this term as "sub" is too easy to mix up with "contained" (i.e. children).

packages/class-core-test/src/gen/LionCore_M3.g.ts
```diff
@@ -286,6 +286,13 @@ export class LionCore_M3Base implements $lwClassCore.ILanguageBase {
     }
 
     public static readonly INSTANCE = new LionCore_M3Base();
+
+    public static isIKeyed(node: $lwClassCore.INodeBase): node is IKeyed {
```
enikao, line 290: I could imagine a generic `isINamed` would also be helpful (for pretty much every language).

dslmeinte: That can only work if we generate for both LionCore and built-ins languages (the `INamed` interface is actually part of the latter), and even then it would only be of limited use: the generated `is<X>` functions can only recognize specializations <X> it knows about, unless we *also* want to do reflection — which is slow as molasses.

packages/class-core-generator/src/helpers/classifiers.ts
dslmeinte: Good point! Done: see latest pushed commit

packages/class-core-test/src/gen/LionCore_M3.g.ts
enikao, line 1: Is there no "normal" language as reference for generator output? Would also ease reviewing the changes in the generator.

packages/class-core-test/src/gen/LionCore_M3.g.ts
dslmeinte: But I could see that it’d be at least useful to make this explicit in the (now-lacking) JSDoc: what do you think?

enikao: Yes, I wouldn't expect that behavior.

dslmeinte: I can also put this behind a feature flag: at least then you have to opt in to this explicitly.

packages/class-core-test/src/gen/LionCore_M3.g.ts
dslmeinte: Would have to be a collection of at least two languages to also check that it works cross-language.

packages/class-core-test/src/gen/LionCore_M3.g.ts
enikao: No idea what's the "natural" expectation in TS context.
My "Yes" referred to "mentioning it in JSDoc".

packages/class-core-test/src/gen/LionCore_M3.g.ts
enikao: Having some test languages to assure generator output seems like a generally good idea, but maybe for a separate PR.

dslmeinte:
I’m shelving this PR for now (by making it a draft), because I think this is more complex than I initially figured, and it’s distracting from things with higher priority. Plus I primarily need this for MyFPL, so I can add a little generator there which doesn’t need to be as generic.

Once I have some time, here’s some ideas:

- Put this behind a feature flag. (Use an object internally to pass everything related to this feature, because we also need to know *all* languages involved in the generation, not just the one for which we’re generating a file.)
- Don’t generate these functions into the language’s base class, but into a separate class in the same file. Documentation saying that the generated type guards only knows about a limited set of languages can then go on the class, instead of on each method.
- Add an `isSuperType(<left classifier>, <right classifier>)` function to `@lionweb/core` which reflectively checks whether <left classifier> is a super type of <right classifier>.

enikao:
Maybe another idea:

Calculate this info at runtime and cache it.
Then reflection might be an option.

dslmeinte:
> Maybe another idea:
> 
> Calculate this info at runtime and cache it. Then reflection might be an option.

Hmmm, you could give this runtime calculation _more_ information than you’d have at generation time, which solves both the information problem (or at least makes it very explicit) and (at least most of) the performance problem. I’m not sure what’s quicker: a string of plain `node instanceof <clazz> ||`, or an object lookup in an array (that’s not really binary searchable). Plus: this can be done generically so we don’t have to extend the generator for it.

So, this is certainly an option.

